Issue #25, opened 2020-03-04
Introduce session.cookie_samesite


roelvanduijnhoven:
### Feature Request

[As of Chrome 80](https://www.chromestatus.com/feature/5088147346030592) a cookie's _SameSite_ setting became more important.

It is however not possible to configure cookie's generated by this repo to set _SameSite_. 

Feature: make it possible to configure using `session.cookie_samesite`.

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    | no

#### Summary

This would require changing [this piece of code](https://github.com/laminas/laminas-session/blob/master/src/SessionManager.php#L442-L450). However, it is not trivial to do so, as setcookie only allows setting this since PHP 7.3.

So to implement this properly for PHP >= 7.0 we need to use setheader explicitly.

roelvanduijnhoven:
> So to implement this properly for PHP >= 7.0 we need to use setheader explicitly.

A quick update. That is not true! We could `SetCookie` from laminas-http to ease the generation :).

PR #30, opened 2020-07-29
Add cookie_samesite parameter

desheffer:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | yes
| RFC           | no
| QA            | no

### Description

Add a `cookie_samesite` parameter, as requested in issue #25.

This parameter corresponds to the `session.cookie_samesite` setting that is [available since PHP 7.3.0](https://www.php.net/manual/en/session.configuration.php#ini.session.cookie-samesite). This setting provides a way to mitigate CSRF (Cross Site Request Forgery) attacks by asserting whether cookies are to be sent along with cross-site requests. It supports three values: `'None'`, `'Lax'`, or `'Strict'`.

### DIFF
--- a/src/Config/ConfigInterface.php
+++ b/src/Config/ConfigInterface.php
@@ -37,6 +37,9 @@ interface ConfigInterface
     public function setCookieDomain($cookieDomain);
     public function getCookieDomain();
 
+    public function setCookieSameSite($cookieSameSite);
+    public function getCookieSameSite();
+
     public function setCookieSecure($cookieSecure);
     public function getCookieSecure();
 
--- a/docs/book/config.md
+++ b/docs/book/config.md
@@ -22,6 +22,7 @@ Option                    | Data Type | Description
 `cookie_httponly`         | `boolean` | Marks the cookie as accessible only through the HTTP protocol.
 `cookie_lifetime`         | `integer` | Specifies the lifetime of the cookie in seconds which is sent to the browser.
 `cookie_path`             | `string`  | Specifies path to set in the session cookie.
+`cookie_samesite`         | `string`  | Specifies whether cookies should be sent along with cross-site requests.
 `cookie_secure`           | `boolean` | Specifies whether cookies should only be sent over secure connections.
 `entropy_length`          | `integer` | Specifies the number of bytes which will be read from the file specified in entropy_file. Removed in PHP 7.1.0.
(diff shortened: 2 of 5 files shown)


desheffer:
@michalbundyra Hi there! I am a new contributor to this repository and I'm not sure how to proceed with getting this merged. To complicate things, coveralls is reporting a coverage decrease, but this is limited to older versions of PHP. I would welcome any feedback or questions you may have. Thanks!

desheffer:
@geerteltink Hi there! I would welcome any feedback or help you could provide on getting this merged. Thanks!

geerteltink:
To get this forward, I would suggest we create another merge request first for #33. This would drop support for < PHP 7.3. This way this build won't fail and we don't need those PHP version checks.

src/Config/ConfigInterface.php
```diff
@@ -37,6 +37,9 @@ public function getCookiePath();
     public function setCookieDomain($cookieDomain);
     public function getCookieDomain();
 
+    public function setCookieSameSite($cookieSameSite);
```
boesing, line 40: adding methods to an interface is a BC break.
We might consider adding a new interfaces.
For example:

- `SameSiteCookieAwareInterface` which extends `SameSiteCookeiCapableInterface`
- `SameSiteCookieCapableInterface` which extends `ConfigInterface`

After this, we can work-around the BC break by checking for these interfaces and trigger exceptions in case someone configures the new configuration parameter but does not implement the new interface.

We can implement the ` in the `SameSiteCookieAwareInterface` in `StandardConfig` and should be good to go.

What do you think?

boesing (changes requested):

boesing:
#35 will be ready soon. After this, you can drop all those PHP 7.3 checks in here as we are always 7.3+

I think I'll retarget this for 2.11.x in case we get around the BC break or 3.0.x in case we want to force a BC break here.

boesing:
I've retargeted this PR against 2.11.x 
Please rebase against that branch as suggested in https://github.com/laminas/laminas-session/pull/30#issuecomment-719941528

desheffer:
Sorry for the delay, I've been busy with other projects. I've removed the special handling for PHP 7.3. I don't have any opinions about the BC break, but I'd be happy to relinquish this PR if someone else has time to explore those options.

PR #35, opened 2020-10-22
PHP 8 support

### LINKED ISSUE #33: PHP 8.0 support
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes

#### Summary

To be prepared for the december release of PHP 8.0, this repository has some additional TODOs to be tested against the new major version.

In order to make this repository compatible, one has to follow these steps:

- [ ] Modify `composer.json` to provide support for PHP 8.0 by adding the constraint `~8.0.0`
- [ ] Modify `composer.json` to drop support for PHP less than 7.3
- [ ] Modify `composer.json` to implement phpunit 9.3 which supports PHP 7.3+
- [ ] Modify `.travis.yml` to ignore platform requirements when installing composer dependencies (simply add `--ignore-platform-reqs` to `COMPOSER_ARGS` env variable)
- [ ] Modify `.travis.yml` to add PHP 8.0 to the matrix (*NOTE*: Do not allow failures as PHP 8.0 has a feature freeze since 2020-08-04!)
- [ ] Modify source code in case there are incompatibilities with PHP 8.0

rieschl:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | yes
| RFC           | no
| QA            | yes

### Description

* The minimum PHP version was bumped to 7.3, PHPUnit updated to 9.3 and all related problems fixed.
* Also, return types were added for test methods and static invocation in assert calls.
* Travis now checks PHP 7.3, 7.4 and 8.0.
  * Code style checks (phpcs) are done in PHP 7.3 because codesniffer v2 throws some deprecation warnings in 7.4. Upgrading to the latest `laminas-coding-standard` will probably fix that.
  * In nightly/PHP8 mongodb is disabled because the extension [doesn't support PHP8](https://github.com/mongodb/mongo-php-driver/issues/1161), yet.
* I also added a test for `Validator\HttpUserAgent`

### ToDo:

- [x] drop `laminas-servicemanager` v2 support
- [x] drop `laminas-eventmanager` v2 support(?)
- [x] remove tests that rely on PHP 7.1 (and are therefore skipped in all builds)
- [x] add some more tests: I'd like to add some more tests, after dropping support for the old laminas-packages

@boesing Could you confirm that the todo list is ok?

fixes #33 
required for #30


composer.json
```diff
@@ -25,7 +25,7 @@
         }
     },
     "require": {
-        "php": "^5.6 || ^7.0",
+        "php": "^7.3 ||^8.0",
```
boesing: Please use `~8.0.0` as described in the initial issue.
We decided in an internal discussion to use this constraint to avoid BC breaks with newer PHP versions.

test/ReflectionUtil.php
```diff
@@ -0,0 +1,23 @@
...
+    /**
+     * @return mixed
+     */
+    public static function getProperty(object $object, string $property)
+    {
+        $reflectionObject   = new ReflectionObject($object);
+        $reflectionProperty = $reflectionObject->getProperty($property);
+        $reflectionProperty->setAccessible(true);
+        $value = $reflectionProperty->getValue($object);
+        $reflectionProperty->setAccessible(false);
```
boesing: You can skip that as the property object wont be used again.

test/ReflectionUtil.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+declare(strict_types=1);
+
+namespace LaminasTest\Session;
+
+use ReflectionObject;
+
+final class ReflectionUtil
```
boesing: Probably change that to a trait and implement in those tests which depend on that.

boesing (changes requested):
So far, LGTM.
Had some minor change requests but overall, nice improvement here 👍

boesing:
> remove tests that rely on PHP 7.1 (and are therefore skipped in all builds)

Do you have examples? I am not into this repo (I've justed added 300 Issues for all packages ;-)) so I'd prefer having some examples. Overall: yes, remove 'em if they're really 7.1 related.

> add some more tests: I'd like to add some more tests, after dropping support for the old laminas-packages

Would prefer a dedicated PR for improvements.

>  drop laminas-servicemanager v2 support
    drop laminas-eventmanager v2 support(?)

I'll give you feedback ASAP.

boesing:
> remove tests that rely on PHP 7.1 (and are therefore skipped in all builds)

Do you have examples? I am not into this repo (I've justed added 300 Issues for all packages ;-)) so I'd prefer having some examples. Overall: yes, remove 'em if they're really 7.1 related.

> add some more tests: I'd like to add some more tests, after dropping support for the old laminas-packages

Would prefer a dedicated PR for improvements.

>  drop laminas-servicemanager v2 support
    drop laminas-eventmanager v2 support(?)

~I'll give you feedback ASAP.~
**Update:** Yes! drop it like its hot! 👍

Ocramius:
Removing old major versions of deps is fine, as long as it doesn't change our API :+1:

rieschl:
> 
> 
> > remove tests that rely on PHP 7.1 (and are therefore skipped in all builds)
> 
> Do you have examples? I am not into this repo (I've justed added 300 Issues for all packages ;-)) so I'd prefer having some examples. Overall: yes, remove 'em if they're really 7.1 related.

Yes, there are some tests which are skipped if PHP is less than `7.1` in [`IdTest`](https://github.com/laminas/laminas-session/blob/2.10.x/test/Validator/IdTest.php#L62),[ `SessionConfigTest`](https://github.com/laminas/laminas-session/blob/2.10.x/test/Config/SessionConfigTest.php#L44) and [`StandardConfigTest`](https://github.com/laminas/laminas-session/blob/2.10.x/test/Config/StandardConfigTest.php#L280)

> 
> > add some more tests: I'd like to add some more tests, after dropping support for the old laminas-packages
> 
> Would prefer a dedicated PR for improvements.

Yes, just a few to keep coveralls from failing :)

boesing:
> Yes, just a few to keep coveralls from failing :)

This PR will get merged even without increasing coveralls. Just decreasing wont be that good ;-)

rieschl:
> 
> 
> > Yes, just a few to keep coveralls from failing :)
> 
> This PR will get merged even without increasing coveralls. Just decreasing wont be that good ;-)

But apparently, coveralls _thinks_ it decreased and marks the build as failed 🤷 
But hopefully, the latest commit will do 🙂

rieschl:
Alright, I think I'm satisfied with my PR, now. However, coveralls still complains and I have no idea, why. If anyone can point me to the right direction, I'm happy to add another test.

boesing:
> However, coveralls still complains and I have no idea, why. 

With the transition to laminas, we've added minimum of coverage which is not reached in this and many other projects. Thus, we are merging those projects even if coveralls is complaining. Thats what I wanted to say in one of my latest messages.

This is already a huge improvement you did here, thank you *very* much, @rieschl 

I'll start review later on, but one thing I've already realized is the deletion of some files.
Even if we dropped support for EventManager v2, dropping source code *is* a BC break.
So either we find a way to keep those files and internally use the v3 event manager or we have to create a new major.
I'd prefer *not* bumping this to a new major. 😅

rieschl:
> I'll start review later on, but one thing I've already realized is the deletion of some files.
> Even if we dropped support for EventManager v2, dropping source code _is_ a BC break.
> So either we find a way to keep those files and internally use the v3 event manager or we have to create a new major.
> I'd prefer _not_ bumping this to a new major. 😅

Hm, I can just keep the files, but then there's really no reason to drop `eventmanger` v2? I could throw an Exception, if [AbstractValidatorChainEM2](https://github.com/laminas/laminas-session/pull/35/commits/a8d0b3f95e3726ca41db5568a106bd8ae5a6410c#diff-b7389785c94b60183a06ae91f15d526e6fdbb90c6805e683e5917d96d8148921) would be loaded?
Or just keep the files but only remove the logic in the [ValidatorChain](https://github.com/laminas/laminas-session/pull/35/commits/a8d0b3f95e3726ca41db5568a106bd8ae5a6410c#diff-c052712441803f7f790b8f646229bcfae9ece43efc14ca7b13cd452f9b4a2fdcL28)?

rieschl:
> 
> 
> > I'll start review later on, but one thing I've already realized is the deletion of some files.
> > Even if we dropped support for EventManager v2, dropping source code _is_ a BC break.
> > So either we find a way to keep those files and internally use the v3 event manager or we have to create a new major.
> > I'd prefer _not_ bumping this to a new major. 😅
> 
> Hm, I can just keep the files, but then there's really no reason to drop `eventmanger` v2? I could throw an Exception, if [AbstractValidatorChainEM2](https://github.com/laminas/laminas-session/pull/35/commits/a8d0b3f95e3726ca41db5568a106bd8ae5a6410c#diff-b7389785c94b60183a06ae91f15d526e6fdbb90c6805e683e5917d96d8148921) would be loaded?
> Or just keep the files but only remove the logic in the [ValidatorChain](https://github.com/laminas/laminas-session/pull/35/commits/a8d0b3f95e3726ca41db5568a106bd8ae5a6410c#diff-c052712441803f7f790b8f646229bcfae9ece43efc14ca7b13cd452f9b4a2fdcL28)?

I reverted the deletion of those files, but kept the new/changed [ValidatorChain](https://github.com/laminas/laminas-session/pull/35/files#diff-e7f831b58cdf0bc67cb2d60bda774d1a2878fb0b9df378bf11b00c5683a6960b). Is this how you meant it?

boesing:
Damn, my fault.
After reviewing this PR locally, I think its fine to drop v2 compat `ValidatorChain`.
I should spend more time in reviewing PRs while checking the code out.

Sorry for the trouble.

rieschl:
> 
> 
> Damn, my fault.
> After reviewing this PR locally, I think its fine to drop v2 compat `ValidatorChain`.
> I should spend more time in reviewing PRs while checking the code out.
> 
> Sorry for the trouble.

Thanks for the feedback!
I removed the latest commit (which re-added the previously deleted files), so the files are gone, now.

rieschl:
I just found another thing. In `SessionConfig` there are [`setHashFunction`](https://github.com/laminas/laminas-session/blob/2.10.x/src/Config/SessionConfig.php#L288) and [`setHashBitsPerCharacter`](https://github.com/laminas/laminas-session/blob/2.10.x/src/Config/SessionConfig.php#L312), and in [`StandardConfig`](https://github.com/laminas/laminas-session/blob/2.10.x/src/Config/StandardConfig.php) even more PHP 7.1 specific stuff.
Now that PHP7.1 is dropped, should I make a no-op there? Removing the methods _is_ a BC break, right? Or just mark it as deprecated, throw an Exception?
Thanks!

boesing:
Just mark the methods as deprecated but keep them as is.
When marked as deprecated, we can drop these functions in the next major and are good to go. 👍

rieschl:
> 
> 
> Just mark the methods as deprecated but keep them as is.
> When marked as deprecated, we can drop these functions in the next major and are good to go. 👍

Done!

boesing (approved):
LGTM

boesing:
@froschdesign As the functionality of some methods in `StandardConfig` was removed with PHP 7.1 (and we dropped 7.1), should we remove the parameters from documentation aswell or should we just mark them as deprecated in the documentation aswell?

boesing:
Thanks, @rieschl!

froschdesign:
@boesing 
> …should we just mark them as deprecated in the documentation aswell?

Has already been done in the past. See: https://docs.laminas.dev/laminas-session/config/#basic-configuration-options

Issue #7, opened 2019-12-31
[Docs] Session Manager initialization code example seems to be severely outdated


weierophinney:
Code and config examples in `Initializing the Session Manager` appear to be in conflict with what is expected by the current code as far as i can see.  In fact, that example is not functional even with fixes suggested in #61 

I'll look into that by tuesday .

---

Originally posted by @Xerkus at https://github.com/zendframework/zend-session/issues/71

weierophinney:
I have the same worry. by the way, if you follow the example and even if you put the correct config array. there is a conflict if you configure the session_name option. the session do not store and is gone! 

---

Originally posted by @ozzpy at https://github.com/zendframework/zend-session/issues/71#issuecomment-287652079

weierophinney:
IMHO custom SessionManager::class factory doesn't fire at all, probably because of an alias in ConfigProvider.php :

`            'aliases' => [
                SessionManager::class => ManagerInterface::class,
            ],`

Some tips needed, plz

---

Originally posted by @duepi at https://github.com/zendframework/zend-session/issues/71#issuecomment-300122297

froschdesign:
The entire code example is useless and wrong. The session manager has a factory and this should be described here.

PR #36, opened 2021-02-09
doc typo: fix config key mismatch

pine3ree:
Later in the example the referred configuration key is `session` instead of `session_manager`.

There is actually a supported `session_manager` configuration key, but the example stands apart and examines this custom configuration by itself. Also, after spotting a `storage` key under `session_manager`, developers may be led to believe that a `storage` key is supported by internal factories (...at least...I know I was).

kind regards

### DIFF
--- a/docs/book/manager.md
+++ b/docs/book/manager.md
@@ -29,7 +29,7 @@ options in your local or global config:
 use Laminas\Session;
 
 return [
-    'session_manager' => [
+    'session' => [
         'config' => [
             'class' => Session\Config\SessionConfig::class,
             'options' => [


froschdesign:
@pine3ree 
> Later in the example the referred configuration key is `session` instead of `session_manager`.

The second example works with a custom configuration structure. The factory for the session manager uses the config key:

https://github.com/laminas/laminas-session/blob/27458b0ee0e6baa2fdb991502a8ea0b703ea56ca/src/Service/SessionManagerFactory.php#L112

But here it is mixed and confusing. The entire page needs an update. Please see #7

froschdesign:
I have an update planned for this week: https://github.com/laminas/laminas-session/milestone/3

pine3ree:
ok @froschdesign,  good! :-)

froschdesign:
@pine3ree 
Your correction is fine, but the whole page is ugly! We need a useful description which helps in the real world and not a copy-paste solution that ignores the factory.

pine3ree:
Hello, @froschdesign. Yes, but I am sorry that I missed that old opened issue of the same nature (outdated docs).

froschdesign:
@pine3ree 
Don't worry about it.

We have some projects for the documentation:

* https://github.com/orgs/laminas/projects/2
* https://github.com/orgs/laminas/projects/3
* https://github.com/orgs/laminas/projects/8

If you have some ideas, please open an issue report in the related issue tracker of the component and we will add it to the projects. Thanks in advance!

PR #46, opened 2021-11-07
Update session bootstrap code to use the same config key used in Initializing the Session Manager

### LINKED ISSUE #2: Fix the configuration index name
Looks like the example code in `bootstrapSession()` expects to read a key `session`and not `session_manager` from the configuration.

Provide a narrative description of what you are trying to accomplish:

- [x] Documentation enhacement

---

Originally posted by @ekarakashi at https://github.com/zendframework/zend-session/pull/124

### LINKED ISSUE #7: [Docs] Session Manager initialization code example seems to be severely outdated

dapphp:
|    Q          |   A
|-------------- | ------
| Documentation | yes

### Description

This is a fix to a code example in the session manager documentation.

The section "Initializing the Session Manager" defines example configuration for the session manager using the key `session_manager`, however, the corresponding code for bootstrapping the session and configuring the session manager are using the key `session` to look for this configuration. This commit updates the code to use to the correct config key `session_manager` instead of `session`.

### DIFF
--- a/docs/book/manager.md
+++ b/docs/book/manager.md
@@ -88,11 +88,11 @@ class Module
         $container->httpUserAgent = $request->getServer()->get('HTTP_USER_AGENT');
 
         $config = $serviceManager->get('Config');
-        if (! isset($config['session'])) {
+        if (! isset($config['session_manager'])) {
             return;
         }
 
-        $sessionConfig = $config['session'];
+        $sessionConfig = $config['session_manager'];
 
         if (! isset($sessionConfig['validators'])) {
             return;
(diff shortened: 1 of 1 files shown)


froschdesign:
@dapphp 
Thanks for your help on this topic! 👍 

> …the corresponding code for bootstrapping the session and configuring the session manager…

All code for the module class should be deleted from the documentation because this code should not be used in an application. And as a demonstration, it is only a deterrent.

Ocramius (approved):
Thanks @dapphp!

Ocramius:
@froschdesign ah, should this be closed instead? 🤔

froschdesign:
The configuration via the module class is a horror and sheds a bad light on this component.
The session manager has a factory and this should be described here. 

I will rewrite the entire page next weekend.

@dapphp 
Thank you for raising the issue! 👍

PR #48, opened 2022-02-14
Clear metadata even if session key didn't exist

### LINKED ISSUE #40: clear() does not clear metadata for provided key
#### Summary
I really do not know if the problem described bellow is a bug or just a layer 8 feature, so I hope this clears the issue a bit.  

#### Current behavior
`\Laminas\Session\Storage\AbstractSessionArrayStorage::clear()` does not remove meta data set with `\Laminas\Session\Storage\AbstractSessionArrayStorage::setMetadata()`

#### How to reproduce
```php
/** 
 * @var \Laminas\Session\Storage\StorageInterface $storage
 */
$storage->setMetadata('foo', 'bar');
$storage->clear('foo');

var_dump($storage->getMetadata('foo'));  // "string" => "bar"
```

#### Expected behavior

```php
/** 
 * @var \Laminas\Session\Storage\StorageInterface $storage
 */
$storage->setMetadata('foo', 'bar');
$storage->clear('foo');

var_dump($storage->getMetadata('foo'));  // "bool" => "false"
```

#### Assumption
The early return statement in the class [AbstractSessionArrayStorage:424](https://github.com/laminas/laminas-session/blob/2.11.x/src/Storage/AbstractSessionArrayStorage.php#L424) prevents the expected behavior. Method `setMetadata()` puts all data into the session index `__Laminas` but `clear()` breaks to early on line `424` but should break after line `432`.

omarkdev:
Session array clear the meta data even if don't exist in array

Close #40

|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### DIFF
--- a/src/Storage/AbstractSessionArrayStorage.php
+++ b/src/Storage/AbstractSessionArrayStorage.php
@@ -436,16 +436,9 @@ abstract class AbstractSessionArrayStorage implements
             return $this;
         }
 
-        if (! isset($_SESSION[$key])) {
-            return $this;
-        }
-
-        // Clear key data
         unset($_SESSION[$key]);
-
-        // Clear key metadata
         $this->setMetadata($key, null)
-             ->unlock($key);
+            ->unlock($key);
 
         return $this;
     }
--- a/test/SessionArrayStorageTest.php
+++ b/test/SessionArrayStorageTest.php
@@ -14,6 +14,9 @@ use function var_export;
  */
 class SessionArrayStorageTest extends TestCase
 {
+    /** @var SessionArrayStorage */
+    private $storage;
+
     protected function setUp(): void
     {
         $_SESSION      = [];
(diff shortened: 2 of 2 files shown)


src/Storage/AbstractSessionArrayStorage.php
```diff
@@ -436,16 +436,14 @@ public function clear($key = null)
             return $this;
         }
 
-        if (! isset($_SESSION[$key])) {
-            return $this;
+        if (isset($_SESSION[$key])) {
```
Ocramius: This conditional is now unnecessary

src/Storage/AbstractSessionArrayStorage.php
```diff
@@ -436,16 +436,14 @@ public function clear($key = null)
...
+            unset($_SESSION[$key]);
         }
 
-        // Clear key data
-        unset($_SESSION[$key]);
-
-        // Clear key metadata
-        $this->setMetadata($key, null)
-             ->unlock($key);
+        if ($this->getMetadata($key)) {
```
Ocramius: I wonder why this check is necessary? 🤔

Ocramius (changes requested):

Ocramius (approved):
Overall LGTM - failures unrelated 

Will ask for further reviews though, because I didn't even know that the session had a concept of metadata :D

samsonasik (approved):

Ocramius:
🚢 as per @samsonasik's review

PR #50, opened 2022-02-15
fix: Implement count in Container

### LINKED ISSUE #17: Implement count() from ArrayObject in Container
At the moment container->count() returns 0 everytime, but it should return the current item count in the container.

---

Originally posted by @bernhard-efler at https://github.com/zendframework/zend-session/issues/52

omarkdev:
Implement count method in Container to retrieve set values quantity

Close #17

|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### DIFF
--- a/src/Container.php
+++ b/src/Container.php
@@ -2,6 +2,8 @@
 
 namespace Laminas\Session;
 
+use function count;
+
 /**
  * Session storage container
  *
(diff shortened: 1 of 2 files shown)


Ocramius (changes requested):
"counting" a session does not make sense, as it is not part of the responsibilities of a session to be "counted".

PR #51, opened 2022-02-21
Prepare for Renovate with reusable workflows

ghostwriter:
|    Q          |   A
|-------------- | ------
| QA            | yes

### Description

Adding Renovate to Laminas, Mezzio repositories. https://gist.github.com/weierophinney/49ca8ae2b916d898dd3d4162b26ff41b

- Configure minimum supported PHP version
- Update composer dependencies
- Update to reusable `Continuous Integration` workflow
- Update to reusable `Automatic Releases` workflow


Ocramius (changes requested):
Same feedback as https://github.com/mezzio/mezzio-swoole/pull/89#pullrequestreview-888416688

Ocramius (approved):

Ocramius:
Hmm, CI is in shambles 🤔

ghostwriter:
Unable to load dynamic library `mongodb.so` on PHP 8.1

weierophinney:
@ghostwriter and @Ocramius — I've fixed the mongodb issue (we had a remnant pre-install operation in here from before we switched PHP 8.1 back to the Sury repo in the CI container).

What I'm seeing now is that when doing `composer update`, we get failures for things we have marked as deprecated, and for which we have `expectDeprecation()` present in the tests. There's a behavior change in PHPUnit sometime between 9.5.9 and 9.5.18 that is causing this, but I haven't been able to isolate it yet. (Tests run fine on 9.5.9 and lower; hoping to figure out the change by bi-secting PHPUnit versions from there.)

weierophinney:
Issue starts in 9.5.10, when `convertDeprecationsToExceptions` was toggled off by default.

Even in 9.5.9 with manually ensuring that flag is toggled on, tests run fine; moving to 9.5.10 causes the same configuration to fail, even though, technically, it should be identical behavior to 9.5.9.

I'll report upstream, but for now, I'll mark those tests skipped.

Ocramius (approved):

Ocramius:
Thanks @ghostwriter and @weierophinney!

PR #54, opened 2022-07-21
Drop support for PHP 7.3

gsteel:
|    Q          |   A
|-------------- | ------
| QA            | yes

### Description

- Adds ocramius/package-versions as allowed plugin (transitive for mongodb during --prefer-lowest)
- Moves service manager to require allowing safe removal of interop container
- bumps dependencies to latest minors
- <del>adds ext-mongodb to dev deps</del>
- Locks `mongodb/mongodb` to `~1.12` because `^1.13` is problematic on PHP 7.4, requiring `ext-mongodb: 1.14` which doesn't appear to be available _(or supported?)_ on 7.4
- Removes ignore-platform-reqs for 8.1
- Applies `strict_types` everywhere with update to LCS 2.3 - This involves changes to arguments sent to `set_cookie` - the methods used to retrieve `HttpOnly` and `Secure` fetch from ini settings, thereby returning "0" or "1", so the method return type docs are lying.

This should enable us to get #53 merged

### DIFF
--- a/src/Config/StandardConfig.php
+++ b/src/Config/StandardConfig.php
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
+
 namespace Laminas\Session\Config;
 
 use Laminas\Session\Exception;
@@ -560,7 +562,7 @@ class StandardConfig implements ConfigInterface, SameSiteCookieCapableInterface
     /**
      * Get session.cookie_secure
      *
-     * @return bool
+     * @return bool|string
      */
     public function getCookieSecure()
     {
@@ -589,7 +591,7 @@ class StandardConfig implements ConfigInterface, SameSiteCookieCapableInterface
     /**
      * Get session.cookie_httponly
      *
-     * @return bool
+     * @return bool|string
      */
     public function getCookieHttpOnly()
     {
--- a/src/Config/StandardConfig.php
+++ b/src/Config/StandardConfig.php
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
-
 namespace Laminas\Session\Config;
 
 use Laminas\Session\Exception;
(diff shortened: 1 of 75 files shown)


src/Config/StandardConfig.php
```diff
@@ -560,7 +562,7 @@ public function setCookieSecure($cookieSecure)
     /**
      * Get session.cookie_secure
      *
-     * @return bool
+     * @return bool|string
```
Ocramius, line 563: Should we even support `string` here? Can't we cast in the return statement?

src/Config/StandardConfig.php
```diff
@@ -589,7 +591,7 @@ public function setCookieHttpOnly($cookieHttpOnly)
     /**
      * Get session.cookie_httponly
      *
-     * @return bool
+     * @return bool|string
```
Ocramius, line 592: Should we even support `string` here? Can't we cast in the return statement?

Ocramius (changes requested):
@gsteel what's your confidence on scalar type hint calls inside this library? Scary or safe?

src/Config/StandardConfig.php
gsteel: The reason I didn't cast in the return statement is due to test failures that verify parity with `assertSame(ini_get('session.whatever'), $foo->getCookieSecure())`

gsteel:
> what's your confidence on scalar type hint calls inside this library? Scary or safe?

Relatively scary. Shall I revert in `/src` and suppress CS so we can get it out the door?

Ocramius:
Yeah, I think that's the safest bet: session data (and input values) can be really scary to deal with.

Let's not introduce instability at this time, and disable the rule for the time being.

Ocramius (approved):
:ship:

PR #56, opened 2022-07-22
Apply PHP 7.4 syntax and typed property

samsonasik:
Signed-off-by: Abdul Malik Ikhsan <samsonasik@gmail.com>

|    Q          |   A
|-------------- | ------
| QA            | yes

### Description

Since composer.json require php 7.4, php 7.4 syntax can be applied with typed properties. 

- For `final class`, typed properties for protected is allowed if no inherit of parent or used by its final class parent.
- Otherwise, update only private properties.

### DIFF
--- a/src/SessionManager.php
+++ b/src/SessionManager.php
@@ -444,7 +444,13 @@ class SessionManager extends AbstractManager
         setcookie(
             $this->getName(), // session name
             '',
-            ['expires' => $_SERVER['REQUEST_TIME'] - 42000, 'path' => $config->getCookiePath(), 'domain' => $config->getCookieDomain(), 'secure' => (bool) $config->getCookieSecure(), 'httponly' => (bool) $config->getCookieHttpOnly()]
+            [
+                'expires' => $_SERVER['REQUEST_TIME'] - 42000,
+                'path' => $config->getCookiePath(),
+                'domain' => $config->getCookieDomain(),
+                'secure' => (bool) $config->getCookieSecure(),
+                'httponly' => (bool) $config->getCookieHttpOnly()
+            ]
         );
     }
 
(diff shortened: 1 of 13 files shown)


src/SessionManager.php
```diff
@@ -446,11 +444,13 @@ public function expireSessionCookie()
...
-            $config->getCookieDomain(),
-            (bool) $config->getCookieSecure(),
-            (bool) $config->getCookieHttpOnly()
+            [
+                'expires' => $_SERVER['REQUEST_TIME'] - 42000, // TTL for cookie
+                'path' => $config->getCookiePath(),
+                'domain' => $config->getCookieDomain(),
+                'secure' => (bool) $config->getCookieSecure(),
+                'httponly' => (bool) $config->getCookieHttpOnly()
+            ]
```
Ocramius: Wait... what happened here? The keys got wrapped/pushed one level down?

samsonasik: that's feature since php 7.3 which can pass it as option for named key https://www.php.net/manual/en/function.setcookie.php 

ref https://github.com/rectorphp/rector/blob/main/docs/rector_rules_overview.md#setcookierector

I can rollback if that's needed.

Ocramius: Yeah, the previous signature is well typed, so I think we **shouldn't** use the `$options` alternative.

The old signature has all parameters type-checked, while the new alternative doesn't seem to have clarity on that, which will be a problem during static analysis.

In fact, `setcookie()` is only superficially inspected in psalm: https://github.com/vimeo/psalm/blob/9b4c8cb53fb5612f81ff632ff945f4a9d6c13ac1/dictionaries/CallMap_73_delta.php#L46

samsonasik: ok, rolled back https://github.com/laminas/laminas-session/pull/56/commits/4580c4575ced29c6653e888fd6f852a2dfc615a8

Ocramius:
Thanks @samsonasik!

PR #60, opened 2022-10-11
Add support for PHP 8.2 to composer.json

TimWolla:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | yes
| RFC           | no
| QA            | no

### Description

Dependency for laminas/laminas-progressbar#17.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,7 +32,7 @@
         }
     },
     "require": {
-        "php": "^7.4 || ~8.0.0 || ~8.1.0",
+        "php": "^7.4 || ~8.0.0 || ~8.1.0 || ~8.2.0",
         "laminas/laminas-eventmanager": "^3.5",
         "laminas/laminas-servicemanager": "^3.15.1",
         "laminas/laminas-stdlib": "^3.10.1"
--- a/.laminas-ci.json
+++ b/.laminas-ci.json
@@ -3,5 +3,8 @@
         "mongodb",
         "mysql",
         "pgsql"
-    ]
+    ],
+    "ignore_php_platform_requirements": {
+        "8.2": true
+    }
 }
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "563df084fce09cb4d254256ad8136a99",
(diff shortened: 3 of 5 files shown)


composer.json
```diff
@@ -32,7 +32,7 @@
         }
     },
     "require": {
-        "php": "^7.4 || ~8.0.0 || ~8.1.0",
+        "php": "^7.4 || ~8.0.0 || ~8.1.0 || ~8.2.0",
```
Ocramius, line 35: Hmm, I think the various `require-dev` will block you here. 

See https://github.com/laminas/laminas-view/blob/69ea122cd53f7839e58cb250975932332e542524/.laminas-ci.json#L7-L10 for an example of how to silence this temporarily.

Ocramius (changes requested):

Ocramius:
Ooof, this will be a pain:

```
E: Unable to locate package php8.2-mongodb
E: Couldn't find any package by glob 'php8.2-mongodb'
```

@boesing WDYT? Can we just skip MongoDB tests, and remove the extension from CI? It's such a massive PITA.

boesing:
should work skipping these tests as of now. 
why does this even need mongo? can we change to apcu probably?

boesing:
Main Problem is that we have fails again - I would prefer having missing packages do not break CI. Tests have to fail when extensions are not available - not container setup. maybe we can modify the container somehow as well.

when adding 8.1, I used shivammathur php-builder which handled plenty of extensions itself. and if not available we just skipped Installation. now this whole setup fails 😔

Ocramius:
Agreed, but we can't really be the integrator for MongoDB when it is such a niche use-case :| 

Having stability for the main use-case is kinda more important here, even if MongoDB then shits the bed

Ocramius:
Still no package released: I really think we should ignore mongodb here, perhaps even deprecate it :thinking:

TimWolla:
FWIW: The laminas-progressbar update was no longer blocked on this and the removal of MongoDB is more than I'm comfortable doing as a contributor. Thus I don't plan to pursue this further.

Feel free to cherry-pick the https://github.com/laminas/laminas-session/commit/479fe973b7dbabad0fbc5cf25240439532d0c4dd.patch commit into 2.14.x (it appears to fail Psalm, though). The remaining changes are trivial anyway.

Ocramius:
Handled by #63 meanwhile. Thanks nonetheless, @TimWolla!

PR #64, opened 2022-11-15
Apply PHP 8.0 Syntax and constructor promotion

samsonasik:
Signed-off-by: Abdul Malik Ikhsan <samsonasik@gmail.com>

|    Q          |   A
|-------------- | ------
| QA            | yes

### Description

Since composer.json require php 8.0, php 8.0 syntax can constuctor promotion can be applied.

### DIFF
--- a/src/Storage/AbstractSessionArrayStorage.php
+++ b/src/Storage/AbstractSessionArrayStorage.php
@@ -67,10 +67,9 @@ abstract class AbstractSessionArrayStorage implements
     /**
      * Get Offset
      *
-     * @param  mixed $key
      * @return mixed
      */
-    public function __get($key)
+    public function __get(mixed $key)
     {
         return $this->offsetGet($key);
     }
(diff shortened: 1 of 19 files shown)


samsonasik:
All green 🎉

src/Storage/AbstractSessionArrayStorage.php
```diff
@@ -67,44 +67,39 @@ public function init($input = null)
     /**
      * Get Offset
      *
-     * @param  mixed $key
      * @return mixed
      */
-    public function __get($key)
+    public function __get(mixed $key)
```
Ocramius, line 72: I don't think this is really valid :thinking: 

I know this was done in an automated way, but the parent interface is much stricter.

Anyway, OK as-is to prevent regressions in downstream implementors.

Ocramius (approved):
Thanks @samsonasik!

PR #79, opened 2024-02-29
Early return from SessionManager::destroy() if headers have already been sent

aleksip:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

In some situations `SessionManager::sessionExists()` may return `true` based on headers having been sent. However, in these cases `session_destroy()` will fail. This change will return early from the `destroy()` method if headers have already been sent and nothing else can be done.

### DIFF
--- a/src/SessionManager.php
+++ b/src/SessionManager.php
@@ -194,11 +194,7 @@ class SessionManager extends AbstractManager
      */
     public function destroy(?array $options = null)
     {
-        if (! $this->sessionExists()) {
-            return;
-        }
-
-        if (headers_sent()) {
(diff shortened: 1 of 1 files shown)


weierophinney (approved):
:ship:

Xerkus:
DO NOT RELEASE

Xerkus:
Session destroy calls session save handler that destroys data server side. Warnings about headers already sent should be supressed.

Ocramius:
@Xerkus this already got released in `2.19.0`

PR #80, opened 2024-02-29
Fix warning in SessionManager::destroy() for headers already sent and actual fix for trying to destroy non-existent session

aleksip:
Fixes problem introduced by #79
 
#79 introduced a fix for trying to destroy non-existent session when headers were already sent. However fix inadvertently prevented valid attempts to destroy session when session is active but headers were already sent.

### DIFF
--- a/src/SessionManager.php
+++ b/src/SessionManager.php
@@ -205,7 +205,7 @@ class SessionManager extends AbstractManager
         }
 
         session_destroy();
-        if ($options['send_expire_cookie']) {
+        if (! headers_sent() && $options['send_expire_cookie']) {
             $this->expireSessionCookie();
         }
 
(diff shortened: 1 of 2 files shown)


src/SessionManager.php
```diff
@@ -194,7 +194,7 @@ protected function initializeValidatorChain()
      */
     public function destroy(?array $options = null)
     {
-        if (headers_sent() || ! $this->sessionExists()) {
+        if (session_status() !== PHP_SESSION_ACTIVE) {
             return;
         }
```
Xerkus, line 201: Let's add inline comment for the change since it is otherwise not enforced by test

```suggestion
        // session_destroy() requires active session while method $this->sessionExists() includes other conditions 
        if (session_status() !== PHP_SESSION_ACTIVE) {
```

Xerkus (approved):

Xerkus:
@aleksip thank you

weierophinney:
Very nice - and looks to be a more thorough solution. Nice work, everyone!

PR #86, opened 2024-06-17
Add the CSRF validator from `laminas-validator`

gsteel:
In validator v3, we'll be upgrading to service manager v4, it's hard dep on `laminas-session` is prohibiting this.

Session, effectively has a hard-dependency on laminas validator and because of the CSRF validator, the validator pkg has a hard dependency on session.

Ideally validator would _not_ depend on session, that's why I'm hoping to deprecate the CSRF validator there in 2.x and remove it, and the dependency on `laminas-session` in 3.x

So, this patch is a lift and shift of the CSRF validator shipped with `laminas-validator`, with some minor type safety improvements.

If there are no objections to this approach, I'll probably need to copy over the documentation for the CSRF validator too.


src/ConfigProvider.php
```diff
@@ -16,6 +17,14 @@ public function __invoke()
...
         return [
             'dependencies' => $this->getDependencyConfig(),
+            'validators'   => [
+                'factories' => [
+                    Validator\Csrf::class => InvokableFactory::class,
+                ],
+                'aliases'   => [
+                    'csrf' => Validator\Csrf::class,
+                ],
+            ],
```
froschdesign: Should be extracted to separate method which can be also used in the `Module` class.

src/Validator/Csrf.php
```diff
@@ -0,0 +1,308 @@
...
+     * @const string
+     */
+    public const NOT_SAME = 'notSame';
+
+    /**
+     * Error messages
+     *
+     * @var array<string, string>
+     */
+    protected array $messageTemplates = [
```
froschdesign, line 42: Do we have rules for visibility in final classes?

froschdesign (changes requested):

src/Validator/Csrf.php
gsteel: I've made everything private except for this variable which is accessed by `AbstractValidator`

gsteel:
@froschdesign - Do you agree that this component is a better place for the CSRF validator?

froschdesign:
@gsteel 
This makes sense to me because we already have session-based validators here and the `CSRF` validator also uses sessions.

mkdocs.yml
```diff
@@ -11,6 +11,7 @@ nav:
         - Introduction: validators/introduction.md
         - "Http User Agent": validators/httpuseragent.md
         - "Remote Address": validators/remoteaddr.md
+        - "CSRF Validator": validator/csrf.md
```
froschdesign: Please use the folder `validators`.

gsteel: Thanks @froschdesign - should be good to go now

froschdesign (approved):

PR #94, opened 2024-11-03
Declare timeout option nullable

hschletz:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | yes

### Description

Fix Psalm complaint when 'timeout' option is disabled (set to NULL - default is 300) via constructor.


gsteel (approved):
Thanks @hschletz

PR #98, opened 2024-12-11
Revert removal of checks for the `SID` constant deprecated in PHP 8.4

### LINKED ISSUE #97: Can not start Session
### Bug Report

|    Q        |   A
|------------ | ------
| Version(s)  | 2.22.0

#### Summary

We are not able to start a Session after setting a Session ID. 

#### Current behavior

The Session does not start

#### How to reproduce

$sessionId = $routeMatch->getParam('sessionId');

$sessionManager->setId($sessionId);
$sessionManager->start();

#### Expected behavior

The Session starts.

#### Possible Cause

The Change in these [lines](https://github.com/laminas/laminas-session/compare/2.21.0...2.22.0#diff-8c351aa956ea03394b47a8a5d217e86224fea5fc21885112746c27ffae7fd81cL97-L98) may cause this.

mimmi20:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

Fixes #97 

### Description

In Release 2.22.0 the check for the Constant SID was removed. This lead to the error that the Session could not be started when a session ID was set.

### DIFF
--- a/src/SessionManager.php
+++ b/src/SessionManager.php
@@ -93,6 +93,9 @@ class SessionManager extends AbstractManager
             return true;
         }
 
+        /**
+         * @var string|false $sid
+         */
         $sid = defined('SID') ? constant('SID') : false;
 
         if ($sid !== false && $this->getId()) {
--- a/src/SessionManager.php
+++ b/src/SessionManager.php
@@ -9,6 +9,8 @@ use Traversable;
 
 use function array_key_exists;
 use function array_merge;
+use function constant;
+use function defined;
 use function headers_sent;
 use function is_array;
 use function iterator_to_array;
@@ -88,6 +90,12 @@ class SessionManager extends AbstractManager
     public function sessionExists()
     {
         if (session_status() === PHP_SESSION_ACTIVE) {
+            return true;
+        }
+
+        $sid = defined('SID') ? constant('SID') : false;
+
+        if ($sid !== false && $this->getId()) {
             return true;
         }
 
--- a/src/SessionManager.php
+++ b/src/SessionManager.php
@@ -91,12 +91,10 @@ class SessionManager extends AbstractManager
             return true;
         }
 
-        if ($this->getId()) {
-            return true;
-        }
         if (headers_sent()) {
             return true;
         }
+
         return false;
     }
 
--- a/test/SessionManagerTest.php
+++ b/test/SessionManagerTest.php
@@ -17,6 +17,7 @@ use Laminas\Session\Storage\SessionStorage;
 use Laminas\Session\Validator\Id;
 use Laminas\Session\Validator\RemoteAddr;
 use LaminasTest\Session\TestAsset\Php81CompatibleStorageInterface;
+use PHPUnit\Framework\Attributes\RunInSeparateProcess;
 use PHPUnit\Framework\TestCase;
 use Traversable;
 
@@ -35,6 +36,7 @@ use function session_start;
 use function session_write_close;
 use function set_error_handler;
 use function stristr;
+use function uniqid;
 use function var_export;
 use function xdebug_get_headers;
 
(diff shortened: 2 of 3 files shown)


mimmi20:
@froschdesign @gsteel There are two tests failing now (`testSessionExistsReturnsTrueWhenSessionStartedThenWritten` and `testCallingWriteCloseShouldNotAlterSessionExistsStatus`).

I dont unserstand why `session_write_close` shall not change the Session-exists-Status.

PHP's page states 

> session_write_close
> (PHP 4 >= 4.0.4, PHP 5, PHP 7, PHP 8)
> 
> session_write_close — Write session data and **end session**

src/SessionManager.php
```diff
@@ -91,12 +91,10 @@ public function sessionExists()
             return true;
         }
 
-        if ($this->getId()) {
-            return true;
-        }
```
gsteel:
```suggestion
        $sid = defined('SID') ? constant('SID') : false;
        if ($sid !== false && $this->getId()) {
            return true;
        }

```

The `SID` constant should be defined if there is an active session. The constant has been deprecated in 8.4 but it is still relevant to apps running any supported version of PHP.

I'm not entirely sure how this is affecting you because the removed check for the `SID` constant should not make any difference anyway.

Can you try patching your installed version of `laminas-session` to revert the lines above and see if your problem persists?

I have tried writing a test case to duplicate your issue locally without success, i.e the following test passes:

```php
    #[RunInSeparateProcess]
    public function testSettingTheIdentifierBeforeStartingTheSessionYieldsTheExpectedId(): void
    {
        $manager = new SessionManager();
        $id = uniqid();
        $manager->setId($id);
        $manager->start();
        $manager->writeClose();

        self::assertTrue($manager->sessionExists());
        self::assertSame($id, $manager->getId());
    }
```

Deleting `if ($this->getId())` here just means that `sessionExists()` will return false when a session does actually exist, hence the failing tests.

What we need is a failing test case that proves the removal of those lines in the last release causes a regression

gsteel (changes requested):

gsteel:
> I dont unserstand why `session_write_close` shall not change the Session-exists-Status.

Closing the session persists the session data for the next request, it doesn't kill the session; There is still an active session once it has been written to storage.

To kill the active session, it must be destroyed with `session_destroy()` or `$manager->destroy()`

test/SessionManagerTest.php
```diff
@@ -902,6 +904,29 @@ public function testIdValidationWillFailOnInvalidData(): void
         $this->manager->start();
     }
 
+    #[RunInSeparateProcess]
```
gsteel, line 828: It's not actually necessary to add this test, there are already tests covering it:

https://github.com/laminas/laminas-session/blob/2.23.x/test/SessionManagerTest.php#L542-L562

This test will pass regardless of whether you've reverted the lines in question or not.

gsteel (changes requested):
The tests that now emit deprecations on 8.4 can be silenced by annotating the test with `#[IgnoreDeprecations]`

test/SessionManagerTest.php
mimmi20: There is actually no test, checking that the session is not marked as started after setting a session id

gsteel: OK - the test will still pass though, regardless of whether those lines in SessionManager have been changed or not.

Can you verify that reverting those lines fixes your problem, and do you have any idea if you're doing something "non-standard" that is causing the problem? If you can isolate that, we can have a new test that fails before, and passes after this change.

mimmi20:
@gsteel Please review again.

gsteel (approved):
Thanks @mimmi20 

I'm going to merge this as-is. Reverting the previously deleted lines in #94 hopefully fixes your problem.

There's still no test here that fails when the `SID` constant is defined and the conditional removed, so I'd appreciate it if you could let us know how/why this fixed your issue.

PR #100, opened 2025-01-10
Check for SaveHandlerInterface in the container and use that if set

TotalWipeOut:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | kinda
| BC Break      | no
| New Feature   | yes
| RFC           | no
| QA            | no

#### My use case
I have a custom laminas session save handler that wraps Googles Firestore save handler. This uses session.save_path for the namespace it uses within the Firestore DB.

so, I configure:
- config option `'session_config' => [ 'save_path' => 'someNamespace']`
- register my custom save handler factory as `Laminas\Session\SaveHandler\SaveHandlerInterface` in ServiceManager as i see that the SessionManager class checks for this and registers it if it exists.

#### The problem
`Laminas\Session\Config\SessionConfig` still requires save_path to exist as a writeable directory, even though the session will start with a `user` save handler due to `SaveHandlerInterface` being registered.
This is because `SessionConfig` class has its own save_handler config option that registers its own custom save handler and only then will it allow save_path not exist locally.
I can't use this config option, as it does not create the save handler from the service manager, and is impossible to inject any dependancies into it. The firestore save handler needs the firestore client injected.

#### Suggested improvement
If a `SaveHandlerInterface`  class is registered in the service manager, and either `'session_config' => 'save_handler' => 'xxx'` is _not_ set, or set to `SaveHandlerInterface::class`, then it will be retrieved from the service container and its instance set as the save_handler. SessionConfig class is already written to handle this, so the only change required is in the factory.

I don't believe it's a BC break, as if its not in the container, it behaves as before. And SessionManager will currently overwrite any save_handler set by SessionConfig if SaveHandlerInterface is set in the service container, so it will not change any other existing configurations or behaviour

### DIFF
--- a/src/Config/SessionConfig.php
+++ b/src/Config/SessionConfig.php
@@ -3,6 +3,7 @@
 namespace Laminas\Session\Config;
 
 use Laminas\Session\Exception;
+use Laminas\Session\SaveHandler\SaveHandlerInterface;
 use SessionHandlerInterface;
 
 use function array_merge;
--- a/src/Config/SessionConfig.php
+++ b/src/Config/SessionConfig.php
@@ -88,8 +88,10 @@ class SessionConfig extends StandardConfig
      * Name of the save handler currently in use. This will either be a PHP
      * built-in save handler name, or the name of a SessionHandlerInterface
      * class being used as a save handler.
+     *
+     * @var null|string|SaveHandlerInterface
      */
-    protected null|string|SaveHandlerInterface $saveHandler = null;
+    protected $saveHandler;
 
     /** @var string session.serialize_handler */
     protected $serializeHandler;
--- a/src/Config/SessionConfig.php
+++ b/src/Config/SessionConfig.php
@@ -87,10 +87,8 @@ class SessionConfig extends StandardConfig
      * Name of the save handler currently in use. This will either be a PHP
      * built-in save handler name, or the name of a SessionHandlerInterface
      * class being used as a save handler.
-     *
-     * @var null|string
      */
-    protected $saveHandler;
+    protected null|string|SaveHandlerInterface $saveHandler = null;
 
     /** @var string session.serialize_handler */
     protected $serializeHandler;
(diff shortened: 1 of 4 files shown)


src/Service/SessionConfigFactory.php
```diff
@@ -55,6 +64,23 @@ public function __invoke(ContainerInterface $container, $requestedName, ?array $
             unset($config['config_class']);
         }
 
+        if (
+            $container->has(SaveHandlerInterface::class)
+            && (! isset($config['save_handler']) || $config['save_handler'] === SaveHandlerInterface::class)
+        ) {
+            $saveHandler = $container->get(SaveHandlerInterface::class);
```
weierophinney: You should have a `has()` check here _first_, so you can throw a more specific error (e.g. "SaveHandlerInterface specified as session save_handler, but the service is not registered in the container").

weierophinney (changes requested):
This feels... off, but looking through the logic of `SessionConfig`, I'm not sure there's another way to do it unless we were to allow specifying known/allowed SaveHandlerInterface implementations outside what we can determine from PHP extensions that register them.

In addition to the comment I've made when retrieving the save handler, please also update the `SessionConfig::$saveHandler` property definition to indicate at type range of `null|string|SaveHandlerInterface`.

src/Service/SessionConfigFactory.php
TotalWipeOut: Thanks for the review!

The `->has()` check is on line 68, as if the service is not registered there is no error, it behaves as before to maintain BC and also this is how SessionManager does it.

I could add an additional check if save_handler is explicitly set to SaveHandlerInterface and then if it is not registered in the service manager, to set the logic to throw in that case?

src/Config/SessionConfig.php
```diff
@@ -87,10 +88,8 @@ class SessionConfig extends StandardConfig
      * Name of the save handler currently in use. This will either be a PHP
      * built-in save handler name, or the name of a SessionHandlerInterface
      * class being used as a save handler.
-     *
-     * @var null|string
      */
-    protected $saveHandler;
+    protected null|string|SaveHandlerInterface $saveHandler = null;
```
samsonasik: this bc break, please only change `@var`

TotalWipeOut: reverted.

src/Service/SessionConfigFactory.php
TotalWipeOut: updated.

TotalWipeOut:
@weierophinney When will you have some time to review my changes? I am needing to use this change in my projects, and just need to understand timescales on this. Thanks as always :)

weierophinney (approved):

PR #102, opened 2025-01-22
Update laminas dependencies

stanis-molnar:
This PR simply updates the laminas dependencies and adapts accordingly the use of the new major versions of laminas-servicemanager and laminas-validator.

I also removed the php annotations and added explicit return types to the modified classes.

This unfortunately breaks backwards compatibility and will require the release of a new major version of laminas-session.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,5 @@
 /.phpcs-cache
-/.phpunit.result.cache
+/.phpunit.cache
 /docs/html/
 /laminas-mkdoc-theme.tgz
 /laminas-mkdoc-theme/
--- a/composer.json
+++ b/composer.json
@@ -34,17 +34,17 @@
     "require": {
         "php": "~8.1.0 || ~8.2.0 || ~8.3.0 || ~8.4.0",
         "laminas/laminas-eventmanager": "^3.12",
-        "laminas/laminas-servicemanager": "^3.22",
+        "laminas/laminas-servicemanager": "^3.22 || ^4.0",
         "laminas/laminas-stdlib": "^3.18"
     },
     "require-dev": {
         "ext-xdebug": "*",
-        "laminas/laminas-cache": "^3.12.2",
-        "laminas/laminas-cache-storage-adapter-memory": "^2.3",
+        "laminas/laminas-cache": "^4.1",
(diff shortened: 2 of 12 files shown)


weierophinney:
@stanis-molnar — I've fixed the Psalm issues that were reported and pushed back to your branch.

I missed the part where you indicated a new major would be needed. I'll check with the @laminas/technical-steering-committee to see how we clear the BC breaks, and verify everyone is okay with the new major.

stanis-molnar:
> @stanis-molnar — I've fixed the Psalm issues that were reported and pushed back to your branch.
> 
> I missed the part where you indicated a new major would be needed. I'll check with the @laminas/technical-steering-committee to see how we clear the BC breaks, and verify everyone is okay with the new major.

@weierophinney  Thanks very much for taking care of that! Please let me know if any further actions on my side are needed.

arhimede:
@weierophinney 
we are about to do 2 different PRs':
- one against V2 to mark all deprecated functions as `deprecated` and to have  new V2 release with that 
- one for V3 in order to implement service manager v4 and remove all `deprecated` functions 

Now with this PR we are confused if is the case or not to commit our PR's, since is a different approach :-) 
Not sure why you say "Since the v2 `createService()` method cannot be removed before v4" , not sure what is the roadmap for this library .

weierophinney:
> @weierophinney we are about to do 2 different PRs':
> 
> * one against V2 to mark all deprecated functions as `deprecated` and to have  new V2 release with that
> * one for V3 in order to implement service manager v4 and remove all `deprecated` functions
> 
> Now with this PR we are confused if is the case or not to commit our PR's, since is a different approach :-) Not sure why you say "Since the v2 `createService()` method cannot be removed before v4" , not sure what is the roadmap for this library .

I wasn't aware you were working on one, @arhimede ; I reviewed this one initially as it looked like it might fit in the 2.23.0 milestone. When I wrote those commit messages, I'd forgotten that this was the v2 series; that should have read "before v3". 

I'm fine with waiting for your PR; marking deprecations first and then doing removal for next major makes sense.

arhimede:
implemented in #103 and the rest of PR's

Issue #23, opened 2019-12-31
Use PHP ext mongodb instead of mongo for v3


weierophinney:
The `Zend\Session\SaveHandler\MongoDB` should use the [new MongoDB extension](http://php.net/manual/en/class.mongodb-driver-manager.php) which runs also under PHP 7 and should not rely on the [deprecated extension mongo](http://php.net/manual/en/class.mongoclient.php) which is not compatible with PHP 7.

I think it should based on the [official MongoDB PHP library](https://github.com/mongodb/mongo-php-library). We can add this library to the suggest Composer section.

What do you think?

---

Originally posted by @sandrokeil at https://github.com/zendframework/zend-session/issues/33

weierophinney:
Did you try https://github.com/alcaeus/mongo-php-adapter?


---

Originally posted by @Maks3w at https://github.com/zendframework/zend-session/issues/33#issuecomment-193424027

weierophinney:
I didn't try it but I think we should update the handler. I can bring a PR if it is accepted.


---

Originally posted by @sandrokeil at https://github.com/zendframework/zend-session/issues/33#issuecomment-193451380

weierophinney:
@sandrokeil - please do! Also, we should definitely target the official wrapper, and not arbitrary third party wrappers.


---

Originally posted by @weierophinney at https://github.com/zendframework/zend-session/issues/33#issuecomment-193453714

weierophinney:
Maybe it's better to not rely on [mongo-php-library](https://github.com/mongodb/mongo-php-library) and use the low level functions of the [MongoDB Driver](http://php.net/manual/de/set.mongodb.php). It's more work, but we have a dependency fewer. I will update my PR depending on the decision.


---

Originally posted by @sandrokeil at https://github.com/zendframework/zend-session/issues/33#issuecomment-202982836

weierophinney:
I prefer remove the support to the old mongo driver and focus the efforts only in the new mongodb extension. But this is not BC,

/fw @weierophinney 


---

Originally posted by @Maks3w at https://github.com/zendframework/zend-session/issues/33#issuecomment-202991180

weierophinney:
@Maks3w ext/mongo has been deprecated, and, more importantly, _does not support PHP 7_. We need to migrate away from it.

Additionally, the page linked by @sandrokeil says specifically:

[...]

PR #103, opened 2025-01-24
Upgrade Service Manager to v4 in `require`, Cache to v4 and Validator to v3 in `require-dev`

### LINKED ISSUE #101: Version 3 + Implement service manager 4
### Feature Request

Implement service manager 4

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    | yes

#### Summary

According to this issue:
https://github.com/laminas/laminas-servicemanager/issues/216
"Only provides servicemanager wiring and thus can add support for v3 & v4" 
so we can give it a try

Jurj-Bogdan:
Upgraded `laminas/laminas-servicemanager` to version 4, as well as the other dependencies required for it,  and added various Psalm tweaks.

I seems I was working on this PR alongside #102 , with the main difference seemingly being the removal of `v2` functions from the codebase in preparation for a new major release.

 I was about to make a different PR to the current branch marking the functions removed here as `deprecated`, but I'll wait for feedback regarding the desired way forward.

### DIFF
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -463,14 +463,6 @@
     </MixedAssignment>
   </file>
   <file src="src/Service/SessionConfigFactory.php">
-    <MixedArgument>
-      <code><![CDATA[$config]]></code>
-      <code><![CDATA[$config['config_class']]]></code>
-      <code><![CDATA[$config['config_class']]]></code>
-    </MixedArgument>
-    <MixedAssignment>
-      <code><![CDATA[$config]]></code>
-    </MixedAssignment>
     <MixedMethodCall>
       <code><![CDATA[new $class()]]></code>
     </MixedMethodCall>
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -463,6 +463,14 @@
     </MixedAssignment>
   </file>
   <file src="src/Service/SessionConfigFactory.php">
+    <MixedArgument>
+      <code><![CDATA[$config]]></code>
+      <code><![CDATA[$config['config_class']]]></code>
+      <code><![CDATA[$config['config_class']]]></code>
+    </MixedArgument>
+    <MixedAssignment>
+      <code><![CDATA[$config]]></code>
+    </MixedAssignment>
     <MixedMethodCall>
       <code><![CDATA[new $class()]]></code>
     </MixedMethodCall>
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -450,13 +450,6 @@
     </PossiblyUnusedReturnValue>
   </file>
   <file src="src/Service/ContainerAbstractServiceFactory.php">
-    <DeprecatedInterface>
-      <code><![CDATA[ContainerAbstractServiceFactory]]></code>
-    </DeprecatedInterface>
-    <MissingConstructor>
-      <code><![CDATA[$config]]></code>
-      <code><![CDATA[$sessionManager]]></code>
-    </MissingConstructor>
     <MixedArgument>
       <code><![CDATA[$config]]></code>
     </MixedArgument>
@@ -466,42 +459,23 @@
     <MixedAssignment>
       <code><![CDATA[$config]]></code>
       <code><![CDATA[$config]]></code>
-      <code><![CDATA[$this->sessionManager]]></code>
+      <code><![CDATA[$sessionManager]]></code>
     </MixedAssignment>
-    <MixedInferredReturnType>
-      <code><![CDATA[null|ManagerInterface]]></code>
-    </MixedInferredReturnType>
-    <MixedReturnStatement>
-      <code><![CDATA[$this->sessionManager]]></code>
-      <code><![CDATA[$this->sessionManager]]></code>
-    </MixedReturnStatement>
-    <ParamNameMismatch>
-      <code><![CDATA[$container]]></code>
-      <code><![CDATA[$container]]></code>
-    </ParamNameMismatch>
-    <RedundantConditionGivenDocblockType>
-      <code><![CDATA[$this->sessionManager !== null]]></code>
-      <code><![CDATA[null !== $this->config]]></code>
-    </RedundantConditionGivenDocblockType>
   </file>
   <file src="src/Service/SessionConfigFactory.php">
-    <DeprecatedInterface>
-      <code><![CDATA[SessionConfigFactory]]></code>
-    </DeprecatedInterface>
+    <MixedArgument>
+      <code><![CDATA[$config]]></code>
+      <code><![CDATA[$config['config_class']]]></code>
+      <code><![CDATA[$config['config_class']]]></code>
+    </MixedArgument>
+    <MixedAssignment>
+      <code><![CDATA[$config]]></code>
+    </MixedAssignment>
     <MixedMethodCall>
       <code><![CDATA[new $class()]]></code>
     </MixedMethodCall>
-    <ParamNameMismatch>
-      <code><![CDATA[$services]]></code>
-    </ParamNameMismatch>
   </file>
   <file src="src/Service/SessionManagerFactory.php">
-    <DeprecatedInterface>
-      <code><![CDATA[SessionManagerFactory]]></code>
-    </DeprecatedInterface>
-    <LessSpecificReturnStatement>
-      <code><![CDATA[$manager]]></code>
-    </LessSpecificReturnStatement>
     <MixedAssignment>
       <code><![CDATA[$config]]></code>
       <code><![CDATA[$configService]]></code>
@@ -510,17 +484,8 @@
       <code><![CDATA[$storage]]></code>
       <code><![CDATA[$validators]]></code>
     </MixedAssignment>
-    <MoreSpecificReturnType>
-      <code><![CDATA[SessionManager]]></code>
-    </MoreSpecificReturnType>
-    <ParamNameMismatch>
-      <code><![CDATA[$services]]></code>
-    </ParamNameMismatch>
   </file>
   <file src="src/Service/StorageFactory.php">
-    <DeprecatedInterface>
-      <code><![CDATA[StorageFactory]]></code>
-    </DeprecatedInterface>
     <MixedArgument>
       <code><![CDATA[$options]]></code>
       <code><![CDATA[$type]]></code>
@@ -530,9 +495,6 @@
       <code><![CDATA[$options]]></code>
       <code><![CDATA[$type]]></code>
     </MixedAssignment>
-    <ParamNameMismatch>
-      <code><![CDATA[$services]]></code>
-    </ParamNameMismatch>
   </file>
   <file src="src/SessionManager.php">
     <DocblockTypeContradiction>
@@ -843,6 +805,11 @@
     <UndefinedDocblockClass>
       <code><![CDATA[CallbackHandler]]></code>
     </UndefinedDocblockClass>
+  </file>
+  <file src="src/Validator/Csrf.php">
+    <MixedArgumentTypeCoercion>
+      <code><![CDATA[$options]]></code>
+    </MixedArgumentTypeCoercion>
   </file>
   <file src="src/Validator/HttpUserAgent.php">
     <PossiblyNullPropertyAssignmentValue>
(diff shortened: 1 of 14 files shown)


arhimede:
@Jurj-Bogdan See https://github.com/laminas/laminas-session/pull/102
Please open a separate PR, which do not break BC , and against 2.23 branch, for the milestone 2.23.1

arhimede:
We have this PR and the other PR, https://github.com/laminas/laminas-session/pull/102
which are doing the same thing.
Only that this PR is removing the deprecated function .

So, should I create the branch 3.0.x and merge this one there ?

gsteel:
> So, should I create the branch 3.0.x and merge this one there

Yes, create 3.0.x from the tip of the next minor release - You should probably get 2.24.x patches resolved first such as #106 so that merge-ups from 2.x have fewer conflicts

arhimede:
@Jurj-Bogdan Please fix the conflicts

src/SaveHandler/Cache.php
```diff
@@ -15,24 +16,18 @@ class Cache implements SaveHandlerInterface
 {
     /**
      * Session Save Path
-     *
-     * @var string
      */
-    protected $sessionSavePath;
+    protected string $sessionSavePath;
```
gsteel: This property is initially null - it is not set in the constructor

src/SaveHandler/Cache.php
```diff
@@ -15,24 +16,18 @@ class Cache implements SaveHandlerInterface
...
 
     /**
      * Session Name
-     *
-     * @var string
      */
-    protected $sessionName;
+    protected string $sessionName;
```
gsteel: This property is initially null - it is not set in the constructor

src/SaveHandler/Cache.php
```diff
@@ -130,32 +117,18 @@ public function gc($maxlifetime)
 
     /**
      * Set cache storage
-     *
-     * @return Cache
      */
-    public function setCacheStorage(CacheStorage $cacheStorage)
+    public function setCacheStorage(CacheStorage $cacheStorage): Cache
```
gsteel: This method should be deprecated and dropped and the constructor changed to use property promotion

src/Service/ContainerAbstractServiceFactory.php
```diff
@@ -41,28 +40,20 @@ class ContainerAbstractServiceFactory implements AbstractFactoryInterface
 {
```
gsteel, line 40: This class should be made final and all properties private. You should also mark this class soft final with `@final` in a 2.x release

src/Service/ContainerAbstractServiceFactory.php
```diff
@@ -41,28 +40,20 @@ class ContainerAbstractServiceFactory implements AbstractFactoryInterface
...
     /**
      * Cached container configuration
-     *
-     * @var array
      */
-    protected $config;
+    protected array $config = [];
```
gsteel: This prop is not set in a constructor, so it's type should be `array|null`

src/Service/ContainerAbstractServiceFactory.php
```diff
@@ -74,62 +65,29 @@ public function canCreate(ContainerInterface $container, $requestedName)
...
     /**
      * Retrieve config from service locator, and cache for later
-     *
-     * @return array
      */
-    protected function getConfig(ContainerInterface $container)
+    protected function getConfig(ContainerInterface $container): array
     {
-        if (null !== $this->config) {
+        if (! empty($this->config)) {
```
gsteel: Please don't use `empty`, the previous `null` check was better

src/Validator/Csrf.php
```diff
@@ -10,27 +10,20 @@
...
 use function is_array;
+use function is_int;
 use function is_string;
 use function md5;
 use function random_bytes;
 use function sprintf;
 use function str_replace;
 use function strtr;
 
-/**
```
gsteel, line 20: Why remove the type declaration? It's useful information that lets the user know what array keys to use.

The required properties can be made non-nullable, such as `name` and `salt`

src/Validator/Csrf.php
```diff
@@ -10,27 +10,20 @@
...
- *     timeout?: ?int,
- * }
- */
 final class Csrf extends AbstractValidator
 {
     /**
      * Error codes
      *
-     * @const string
+     * @var string
```
gsteel: Doc blocks like this on constants can just be removed. They're not helpful

src/Validator/Csrf.php
```diff
@@ -46,33 +39,50 @@ final class Csrf extends AbstractValidator
...
-    private ?Container $session = null;
+    private ?Container $session;
 
     /**
      * TTL for CSRF token
      */
-    private int|null $timeout = 300;
+    private ?int $timeout;
 
-    /** @param OptionsArgument $options */
```
gsteel, line 72: Please revert as per comment on class-level doc-block

src/Validator/Csrf.php
```diff
@@ -10,27 +10,20 @@
...
 
-/**
- * @psalm-type OptionsArgument = array{
- *     name?: non-empty-string,
- *     salt?: non-empty-string,
- *     session?: Container,
- *     timeout?: ?int,
- * }
- */
 final class Csrf extends AbstractValidator
```
gsteel, line 37: This whole class will probably need further refactoring such as deprecating and dropping all the public setters and getters

test/Service/SessionManagerFactoryTest.php
```diff
@@ -114,7 +114,8 @@ public function testFactoryWillAddValidatorViaConfiguration(): void
 
         $manager->start();
 
-        $chain     = $manager->getValidatorChain();
+        $chain = $manager->getValidatorChain();
+        /** @psalm-suppress ArgumentTypeCoercion **/
```
gsteel: Please remove this suppression and replace it with assertions.

Assert the manager is `ManagerInterface`
Assert the chain is a `EventManagerInterface` or whatever it's supposed to be

test/Service/SessionManagerFactoryTest.php
```diff
@@ -198,7 +199,8 @@ public function testFactoryDoesNotAttachValidatorTwoTimes(): void
             // Ignore exception, because we are not interested whether session validation passes in this test
         }
 
-        $chain     = $manager->getValidatorChain();
+        $chain = $manager->getValidatorChain();
+        /** @psalm-suppress ArgumentTypeCoercion **/
```
gsteel: Same here… More assertions!

psalm-baseline.xml
```diff
@@ -1019,6 +963,11 @@
       <code><![CDATA[CallbackHandler]]></code>
     </UndefinedDocblockClass>
   </file>
+  <file src="src/Validator/Csrf.php">
```
gsteel: We really shouldn't be expanding the baseline while moving towards 3 - This is likely due to dropping the defined array shape on the CSRF validator

gsteel (changes requested):

src/Validator/Csrf.php
gsteel: Not suggesting that work should happen in this patch tho!

src/Validator/Csrf.php
Jurj-Bogdan: Reverted this change, but I also had to add the options passed on to `AbstractValidator`'s constructor to avoid psalm errors - this however feels weird to me, as any change in the parent package will cause issues here, which is why I removed the annotation initially.

test/Service/SessionManagerFactoryTest.php
Jurj-Bogdan: Added some phpunit assertions as they're useful, but Psalm is still adamant in throwing "ArgumentTypeCoercion" errors which is why the suppressions are still there.

src/Service/ContainerAbstractServiceFactory.php
Jurj-Bogdan: Done - also a related question here. I just noticed there are already some classes marked as deprecated, maybe this should be a good time to remove them as well? 

Talking about `AbstractValidatorChain` and `AbstractValidatorChainEM3`

src/Validator/Csrf.php
froschdesign: > …but I also had to add the options passed on to `AbstractValidator`'s constructor to avoid psalm errors…

Right, but this is the way it is and it is still better than not giving the user any information.
Compare with:

https://github.com/laminas/laminas-validator/blob/b0b640aa6c967ba53e2d18150994bd1e1903c678/src/Uri.php#L14-L25

src/Service/ContainerAbstractServiceFactory.php
```diff
@@ -74,51 +65,20 @@ public function canCreate(ContainerInterface $container, $requestedName)
...
-        return $this($container, $requestedName);
-    }
-
     /**
      * Retrieve config from service locator, and cache for later
-     *
-     * @return array
      */
-    protected function getConfig(ContainerInterface $container)
+    protected function getConfig(ContainerInterface $container): array
```
gsteel: This should be `private`

src/Service/ContainerAbstractServiceFactory.php
```diff
@@ -143,29 +103,28 @@ protected function getConfig(ContainerInterface $container)
 
     /**
      * Retrieve the session manager instance, if any
-     *
-     * @return null|ManagerInterface
      */
-    protected function getSessionManager(ContainerInterface $container)
+    protected function getSessionManager(ContainerInterface $container): ?ManagerInterface
```
gsteel: This should be `private`

src/Service/ContainerAbstractServiceFactory.php
```diff
@@ -143,29 +103,28 @@ protected function getConfig(ContainerInterface $container)
...
     }
 
     /**
      * Normalize the container name in order to perform a lookup
-     *
-     * @param  string $name
-     * @return string
      */
-    protected function normalizeContainerName($name)
+    protected function normalizeContainerName(string $name): string
```
gsteel: This should be `private` or maybe inlined if it's only used once

test/Service/SessionManagerFactoryTest.php
gsteel: You need to assert `$chain` is an instance of `EventManager` (The concrete class) because that's what `getListenersForEvent` expects.

Please also add `findUnusedPsalmSuppress="true"` to `psalm.xml` so that unused  inline suppressions are flagged.

test/Service/SessionManagerFactoryTest.php
gsteel: This suppression still needs fixing too!

gsteel (changes requested):
Thanks @Jurj-Bogdan - A few more tweaks please :)

Apart from inline comments, all of the changed factories need to be made `final` and they should all have a good cleanup - i.e `protected` props to `private`, drop any methods that are pointless etc. Happy to wait for another patch for those things, but we really don't want users extending factories and need to prevent that in the next major

src/Service/ContainerAbstractServiceFactory.php
gsteel: Yes - _anything_ that's deprecated should be dealt with in the next major. I don't know this lib well enough to know what the implications or plans might have been though.

Don't forget to document any BC breaks here with companion patches against 2.x including deprecations, and, make a start on a migration guide that lists all the BC breaking changes

Jurj-Bogdan:
@gsteel Just added the latest requested changes, i hope i didn't miss anything. A note related to the `Csrf.php` changes however:

My understanding of the comment is that `Csrf` objects should only be initialised via constructor, with all properties being immutable - is this correct? 

Since the old getters for both `hash` and `session` set values for each of them in case they were null, i went ahead and made them not nullable in the first place, although the the changes in the constructor might be a bit ugly. 

Also removed the sequences marked as "@todo remove", including the adding of the hash to the session container directly instead of as an entry in the "tokenList" only - i take it this feature is no longer wanted, right?

As for the failing test, since issue #115 mentions removing mongoDb altogether, I'll go ahead and create a different PR with that change (and for the other sub issues as well) so this PR doesn't get too cluttered.

gsteel:
If all the red CI here is MongoDB's fault, why not work on a patch to fix #115 first and then rebase here when its done.

gsteel:
Mind you, looking at the CI logs, the problem is a simple deprecation in the mongo lib. Has anyone looked at just fixing that deprecation?

Jurj-Bogdan:
> Mind you, looking at the CI logs, the problem is a simple deprecation in the mongo lib. Has anyone looked at just fixing that deprecation?

That would be the fix in another PR for the 2.x version, but from what i understand mongoDb is to be removed in version 3, no?

gsteel:
> That would be the fix in another PR for the 2.x version, but from what i understand mongoDb is to be removed in version 3, no?

There's currently no rationale for its removal as far as I know - just an issue to remove it.

Perhaps @arhimede has more information on why it should be removed

arhimede:
> > That would be the fix in another PR for the 2.x version, but from what i understand mongoDb is to be removed in version 3, no?
> 
> There's currently no rationale for its removal as far as I know - just an issue to remove it.
> 
> Perhaps @arhimede has more information on why it should be removed

First of all, MongoDB is an external dependency of this package , I see no point to be included in this package .
When there is a need , a special laminas-session-mongodb package can be created .
I totally agree with your comment here : https://github.com/laminas/laminas-session/issues/115#issuecomment-2714841397

Last but not least, see this old discussion on this matter: https://github.com/laminas/laminas-session/issues/23

arhimede:
And not to mention the licensing of MongoDB which is quite complex.

gsteel:
I have no problem with it being removed as long as the decision making and docs are sound 👍 

My point here is: Why not send in a patches now to…

- Deprecate mongo features in 2.x
- Delete them in 3.x

That way devlopment can move forward on other things more easily because there will be fewer red builds

Jurj-Bogdan:
Rebased to 3.0.x, and since it was already bumped to version `3.0` i also moved `laminas/laminas-validator` to `require`.

arhimede:
@laminas/technical-steering-committee  Can somebody take a look at this PR ?

src/Validator/Csrf.php
```diff
@@ -46,33 +53,56 @@ final class Csrf extends AbstractValidator
...
+        $name    = $options['name'] ?? 'csrf';
+        $salt    = $options['salt'] ?? 'salt';
+        $session = $options['session'] ?? null;
+        $timeout = $options['timeout'] ?? 300;
+
+        assert(is_string($hash) || $hash === null);
+        assert(is_string($name) && $name !== '');
+        assert(is_string($salt) && $salt !== '');
+        assert($session instanceof Container || $session === null);
+        assert(is_int($timeout) || $timeout === null);
```
Ocramius: `$timeout` is never `null`?

src/Validator/Csrf.php
```diff
@@ -46,33 +53,56 @@ final class Csrf extends AbstractValidator
...
+    private string $salt;
 
-    private ?Container $session = null;
+    private Container $session;
 
     /**
      * TTL for CSRF token
      */
-    private int|null $timeout = 300;
+    private ?int $timeout;
```
Ocramius: This is never `null` now. would suggest making it a `positive-int` too

Ocramius (approved):
Some tiny changes needed, otherwise :ship:

PR #120, opened 2025-03-21
Replace Laminas\Cache requirement with PSR-16 Cache

### LINKED ISSUE #112: replace laminas-cache with a PSR interface
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    | yes

#### Summary

Remove the dependency to laminas-cache 
And replace it with a PSR interface    PSR-6

Jurj-Bogdan:
|    Q          |   A
|-------------- | ------
| BC Break      | yes
| RFC           | yes

### Description

Replaced `laminas/laminas-cache` with `psr/simple-cache` in the `Cache` SaveHandler as discussed in issue #112.

This is however an initial version, as i have a few questions related to the `Cache` class, which i'll go over in comments on the specific lines. I'll go ahead and add the relevant documentation as soon as a way forward is set.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -39,17 +39,15 @@
     },
     "require-dev": {
         "ext-xdebug": "*",
-        "laminas/laminas-cache": "^3.12.2",
-        "laminas/laminas-cache-storage-adapter-memory": "^2.3",
         "laminas/laminas-coding-standard": "~3.0.1",
         "laminas/laminas-http": "^2.20",
         "laminas/laminas-validator": "^2.64.1",
         "phpunit/phpunit": "^10.5.38",
         "psalm/plugin-phpunit": "^0.19.0",
+        "psr/simple-cache": "^3.0.0",
         "vimeo/psalm": "^5.26.1"
     },
     "suggest": {
-        "laminas/laminas-cache": "Laminas\\Cache component",
         "laminas/laminas-http": "Laminas\\Http component",
         "laminas/laminas-servicemanager": "Laminas\\ServiceManager component",
         "laminas/laminas-validator": "Laminas\\Validator component"
--- a/composer.json
+++ b/composer.json
@@ -35,7 +35,8 @@
         "php": "~8.1.0 || ~8.2.0 || ~8.3.0 || ~8.4.0",
         "laminas/laminas-eventmanager": "^3.12",
         "laminas/laminas-servicemanager": "^3.22",
-        "laminas/laminas-stdlib": "^3.18"
+        "laminas/laminas-stdlib": "^3.18",
+        "psr/simple-cache": "^3.0"
     },
     "require-dev": {
         "ext-xdebug": "*",
@@ -44,10 +45,10 @@
         "laminas/laminas-validator": "^2.64.1",
         "phpunit/phpunit": "^10.5.38",
         "psalm/plugin-phpunit": "^0.19.0",
-        "psr/simple-cache": "^3.0.0",
         "vimeo/psalm": "^5.26.1"
     },
     "suggest": {
+        "psr/simple-cache-implementation": "Find implementations of PSR-16 in this virtual package",
         "laminas/laminas-http": "Laminas\\Http component",
         "laminas/laminas-servicemanager": "Laminas\\ServiceManager component",
         "laminas/laminas-validator": "Laminas\\Validator component"
--- a/docs/book/v2/save-handler.md
+++ b/docs/book/v2/save-handler.md
@@ -5,36 +5,6 @@ are decoupled from PHP's save handler functions and are only implemented as a
 PHP save handler when utilized in conjunction with
 `Laminas\Session\SessionManager`.
 
-## Cache
-
-`Laminas\Session\SaveHandler\Cache` allows you to provide an instance of
-`Laminas\Cache\Storage\Adapter\AdapterInterface` to be utilized as a session save
-handler. Generally if you are utilizing the `Cache` save handler; you are likely
-using products such as memcached.
-
-### Basic usage
-
-A basic example is one like the following:
-
-```php
-use Laminas\Cache\StorageFactory;
-use Laminas\Session\SaveHandler\Cache;
-use Laminas\Session\SessionManager;
-
-$cache = StorageFactory::factory([
-    'adapter' => [
-       'name' => 'memcached',
-       'options' => [
-           'server' => '127.0.0.1',
-       ],
-    ],
-]);
-
-$saveHandler = new Cache($cache);
-$manager = new SessionManager();
-$manager->setSaveHandler($saveHandler);
-```
-
 ## DbTableGateway
 
 `Laminas\Session\SaveHandler\DbTableGateway` allows you to utilize
--- a/docs/book/v2/save-handler.md
+++ b/docs/book/v2/save-handler.md
@@ -5,6 +5,36 @@ are decoupled from PHP's save handler functions and are only implemented as a
 PHP save handler when utilized in conjunction with
 `Laminas\Session\SessionManager`.
 
+## Cache
+
+`Laminas\Session\SaveHandler\Cache` allows you to provide an instance of
+`Laminas\Cache\Storage\Adapter\AdapterInterface` to be utilized as a session save
+handler. Generally if you are utilizing the `Cache` save handler; you are likely
+using products such as memcached.
+
+### Basic usage
+
+A basic example is one like the following:
+
+```php
+use Laminas\Cache\StorageFactory;
+use Laminas\Session\SaveHandler\Cache;
+use Laminas\Session\SessionManager;
+
+$cache = StorageFactory::factory([
+    'adapter' => [
+       'name' => 'memcached',
+       'options' => [
+           'server' => '127.0.0.1',
+       ],
+    ],
+]);
+
+$saveHandler = new Cache($cache);
+$manager = new SessionManager();
+$manager->setSaveHandler($saveHandler);
+```
+
 ## DbTableGateway
 
 `Laminas\Session\SaveHandler\DbTableGateway` allows you to utilize
--- a/src/SaveHandler/Cache.php
+++ b/src/SaveHandler/Cache.php
@@ -3,31 +3,17 @@
 namespace Laminas\Session\SaveHandler;
 
 use Psr\SimpleCache\CacheInterface;
-use ReturnTypeWillChange;
 
 /**
  * Cache session save handler
- *
- * @see ReturnTypeWillChange
  */
-class Cache implements SaveHandlerInterface
+final class Cache implements SaveHandlerInterface
 {
-    /**
-     * Session Save Path
-     */
-    protected string $sessionSavePath;
-
-    /**
-     * Session Name
-     */
-    protected string $sessionName;
-
-    /**
-     * Constructor
-     */
-    public function __construct(protected CacheInterface $cacheStorage)
-    {
-        $this->setCacheStorage($cacheStorage);
+    public function __construct(
+        private CacheInterface $cacheStorage,
+        private string $sessionSavePath,
+        private string $sessionName
+    ) {
     }
 
     /**
@@ -35,7 +21,6 @@ class Cache implements SaveHandlerInterface
      */
     public function open(string $path, string $name): bool
     {
-        // @todo figure out if we want to use these
         $this->sessionSavePath = $path;
         $this->sessionName     = $name;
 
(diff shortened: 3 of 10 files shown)


src/SaveHandler/Cache.php
```diff
@@ -15,42 +14,26 @@ class Cache implements SaveHandlerInterface
...
-     *
-     * @param string $path
-     * @param string $name
-     * @return bool
      */
-    #[ReturnTypeWillChange]
-    public function open($path, $name)
+    public function open(string $path, string $name): bool
     {
         // @todo figure out if we want to use these
```
Jurj-Bogdan: How do you suggest handling this feature? And on a related note, should I go ahead and set the `path` and `name` in the constructor or set default values?

src/SaveHandler/Cache.php
```diff
@@ -61,101 +44,62 @@ public function open($path, $name)
...
      * Read session data
-     *
-     * @param string $id
-     * @return string
      */
-    #[ReturnTypeWillChange]
-    public function read($id)
+    public function read(string $id): string
     {
-        return (string) $this->getCacheStorage()->getItem($id);
```
Jurj-Bogdan, line 82: I noticed that `read` was made to always return strings, which i've kept for now. Would it not be useful however to return false if nothing was read, as in `SessionHandler`?

src/SaveHandler/Cache.php
```diff
@@ -61,101 +44,62 @@ public function open($path, $name)
...
 
     /**
      * Garbage Collection
-     *
-     * @param int $maxlifetime
-     * @return bool
      */
-    #[ReturnTypeWillChange]
-    public function gc($maxlifetime)
+    public function gc(int $maxlifetime): bool
```
Jurj-Bogdan: i'd also need some feedback for this function as well - psr-16 provides no method to clear items older than the provided `maxlifetime`, nor any way to get them either.

gsteel: Why PSR-16 as opposed to PSR-6? Just a question rather than a reccommendation. I've never really used simple cache.

cc @laminas/technical-steering-committee 

Edit: PSR-6 enables you to differentiate between null cache values meaning there is no cached item vs null values meaning null was cached. This might be relevant when determining if a session exists or not.

src/SaveHandler/Cache.php
gsteel: In the constructor IMO. Path and name are configuration details

gsteel: This has to match PHP's built-in interface, so it'll have to stay as-is probably

src/SaveHandler/Cache.php
gsteel: Returning `string|false` matches PHP's built-in session handler interface

src/SaveHandler/Cache.php
```diff
@@ -61,101 +44,62 @@ public function open($path, $name)
...
-
-    /**
-     * @deprecated Misspelled method - use getCacheStorage() instead. Will be removed in version 3.0
-     *
-     * @return CacheStorage
-     */
-    public function getCacheStorge()
-    {
-        return $this->getCacheStorage();
-    }
```
gsteel: Cache setters and getters need to get in the bin

src/SaveHandler/Cache.php
```diff
@@ -15,42 +14,26 @@ class Cache implements SaveHandlerInterface
...
-     * @var CacheStorage
-     */
-    protected $cacheStorage;
+    protected string $sessionName;
 
     /**
      * Constructor
      */
-    public function __construct(CacheStorage $cacheStorage)
+    public function __construct(protected CacheInterface $cacheStorage)
```
gsteel: We should make this class final and all properties private IMO

src/SaveHandler/Cache.php
Jurj-Bogdan: > Why PSR-16 as opposed to PSR-6? Just a question rather than a reccommendation. I've never really used simple cache.
> 
> cc @laminas/technical-steering-committee
> 
> Edit: PSR-6 enables you to differentiate between null cache values meaning there is no cached item vs null values meaning null was cached. This might be relevant when determining if a session exists or not.

I was looking at PSR-6 initially as well, but since it was mentioned in #112 I went with PSR-16 as it feels that the simpler interface would fit this package's usage well enough and the users could have more lightweight implementations in the end. Would the feature you mentioned warrant a change to PSR6?
 
As for the `gc` method here, just uploaded all the other review points with this left. Since neither PSR-6 or PSR-16 require a method that explicitly fits this case, it's pretty much left to implement a version here.

Correct me if i'm wrong but that would probably be made by checking the session files in the specified `sessionSavePath` and removing any unchanged older than the `maxlifetime` sent to the method - would that be overkill for this package? Or should users be left to implement a `gc` that fits their use case?

src/SaveHandler/Cache.php
Jurj-Bogdan: In this case I'll also add a PR for the 2.x version marking it as `@final`.

src/SaveHandler/Cache.php
gsteel: The point of `gc` is to iterate over all sessions and destroy those older than `$maxlifetime`. Neither PSR-6 nor PSR-16 will help you do that - you'd need to maintain a list of sessions to map and update timestamps, which itself would be a waste of time, because the items would naturally expire based on the cache TTL. I think that `gc` should probably just be a no-op.

WRT PSR-6 vs PSR-16 - Can we possibly support both?

i.e. `__construct(CacheItemPoolInterface|CacheInterface $cache)`

[`mezzio-session-cache`](https://github.com/mezzio/mezzio-session-cache) is PSR-6 only

I don't personally mind what gets chosen, but as a general purpose session save handler for MVC stuff, using a PSR is defo a thumbs up, but forcing a choice between 6 or 16 just seems off to me.

Packagist install stats are even for both PSRs

It'd be good to have more opinions here

composer.json
```diff
@@ -39,19 +39,17 @@
...
-        "laminas/laminas-cache": "^3.12.2",
-        "laminas/laminas-cache-storage-adapter-memory": "^2.3",
         "laminas/laminas-coding-standard": "~3.0.1",
         "laminas/laminas-db": "^2.20.0",
         "laminas/laminas-http": "^2.20",
         "laminas/laminas-validator": "^2.64.1",
         "mongodb/mongodb": "~1.20.0",
         "phpunit/phpunit": "^10.5.38",
         "psalm/plugin-phpunit": "^0.19.0",
+        "psr/simple-cache": "^3.0.0",
```
gsteel: This should be in `require` - its symbols are used in `/src`

composer.json
```diff
@@ -39,19 +39,17 @@
...
         "vimeo/psalm": "^5.26.1"
     },
     "suggest": {
-        "laminas/laminas-cache": "Laminas\\Cache component",
```
gsteel, line 52: We should suggest `psr/simple-cache-implementation` and/or `psr/cache-implementation` here

src/SaveHandler/Cache.php
```diff
@@ -2,57 +2,25 @@
...
-     *
-     * @param string $path
-     * @param string $name
-     * @return bool
      */
-    #[ReturnTypeWillChange]
-    public function open($path, $name)
+    public function open(string $path, string $name): bool
     {
-        // @todo figure out if we want to use these
```
gsteel, line 55: Not sure this method needs to do anything other than `return true` - there is no "open" step, we just read and write to the cache

src/SaveHandler/Cache.php
```diff
@@ -2,57 +2,25 @@
...
-    /**
-     * Constructor
-     */
-    public function __construct(CacheStorage $cacheStorage)
-    {
-        $this->setCacheStorage($cacheStorage);
+    public function __construct(
+        private CacheInterface $cacheStorage,
+        private string $sessionSavePath,
+        private string $sessionName
```
gsteel: Not sure we need these properties at all

src/SaveHandler/Cache.php
gsteel: You're returning `mixed|false` here

i.e.

```php
if (! $this->cacheStorage->has($id)) {
    return false;
}

return (string) $this->cacheStorage->get($id, '');
```

gsteel (changes requested):

src/SaveHandler/Cache.php
Jurj-Bogdan: > WRT PSR-6 vs PSR-16 - Can we possibly support both?
> 
> i.e. `__construct(CacheItemPoolInterface|CacheInterface $cache)`
> 
> [`mezzio-session-cache`](https://github.com/mezzio/mezzio-session-cache) is PSR-6 only
> 
> I don't personally mind what gets chosen, but as a general purpose session save handler for MVC stuff, using a PSR is defo a thumbs up, but forcing a choice between 6 or 16 just seems off to me.
> 
> Packagist install stats are even for both PSRs
> 
> It'd be good to have more opinions here

Haven't touched this point in the latest commit - thinking about it, supporting both PSRs does make sense though if that's what's decided.

 But since the interfaces provide different methods, maybe it would be better as two separate files? Maybe renaming the current `Cache` class to `SimpleCache` and a new `PSR6Cache` that uses the PSR6 interface

gsteel: That's a good idea. A separate save handler for each of PSR6 and 16 would be better than shoe-horning support for both into a single class!

Lets just choose and stick with 16 for this patch and class and then implement an adapter for PSR-6 in another patch.

gsteel:
This LTGM @Jurj-Bogdan

Before I approve for merge, the other PRs with Mongo and Laminas DB need merging first, followed by a rebase here.

docs/book/v2/save-handler.md
```diff
@@ -5,36 +5,6 @@ are decoupled from PHP's save handler functions and are only implemented as a
...
-           'server' => '127.0.0.1',
-       ],
-    ],
-]);
-
-$saveHandler = new Cache($cache);
-$manager = new SessionManager();
-$manager->setSaveHandler($saveHandler);
-```
-
```
gsteel: This needs to be reverted. The v2 docs should remain unchanged, or, amended to include a note such as

```markdown
CAUTION: **Some Headline**
In version 3, `laminas-session` will no longer support `laminas-cache` as a storage backend.
Instead, a PSR-16 cache will be required.
You can find out more in the [migration guide](../v3/migration.md).
```

To be clear, just reverting the deletion is fine IMO, because the migration guide should have all the relevant info.

gsteel (changes requested):
V3 docs for using a PSR-16 cache implementation are not done yet - Are you going to do these in a separate patch or here?

If you choose a separate patch, can you add an issue for it so it doesn't get forgotten.

Looking at the [PSR-16 Spec](https://www.php-fig.org/psr/psr-16/#12-definitions) for cache keys, are we confident that the session ids we receive will validate as per the spec. If we're not confident of that, we would probably need to use a pre-determined key hash as the internal cache key, for example:

```php
public function read(string $id): string|false
{
    $cacheKey = hash('xxh32', $id);
    // ...
}
```

Additionally, nowhere are we catching PSR cache exceptions and re-throwing them in our own namespace.

Assuming we don't hash the keys to a known dictionary, and cannot guarantee that the cache impl. won't throw, we should be doing something like:

```php
/**
 * @throws \Laminas\Session\WhateverException For Reasons…
 */
public function read(string $id): string|false
{
    try {
        return $this->cache->get($id);
    } catch (PSRException $e) {
        throw WhateverException::becauseTheKeyWasBad($e);
    }
}
```

The test case for the handler verifies interactions with a cache impl well enough with mocks, but I would like to see an integration test with an actual implementation. This also serves as reference material for docs. i.e. we can dog food laminas cache here (Assuming it has adapters that implement PSR-16). I won't die on that hill, but arguably its easier to write the integration test than the mocked version.

As always, thanks for your work @Jurj-Bogdan

Jurj-Bogdan:
Just added a few checks without mocking, although i'm open to suggestions about any additions - also reverted the documentation which I must have deleted by mistake

src/SaveHandler/Cache.php
```diff
@@ -2,160 +2,89 @@
...
-     *
-     * @var CacheStorage
-     */
-    protected $cacheStorage;
-
-    /**
-     * Constructor
-     */
-    public function __construct(CacheStorage $cacheStorage)
+    public function __construct(private CacheInterface $cacheStorage)
```
gsteel:
```suggestion
    public function __construct(private readonly CacheInterface $cacheStorage)
```

gsteel (changes requested):
`Cache::read`, `Cache::write` and `Cache::destroy` need test coverage that asserts the expected exception is thrown when the underlying cache adater crashes. Also, we should assert that a known invalid cache key given to read, write or destroy will not cause an exception because the keys are hashed.

gsteel (approved):
Thanks @Jurj-Bogdan 💪

Issue #114, opened 2025-03-05
define laminas-validator as a hard requirement


arhimede:
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    | yes

#### Summary

Define laminas-validator as a hard requirement, in order to be sure config files are valid

marcelthole:
Could be closed. See #103 (And the duplicates already #126 #127 )

arhimede:
Implemented in #103

PR #126, opened 2025-04-25
Move laminas-validator to require dependencies and bump the version with all dependencies

marcelthole:
|    Q          |   A
|-------------- | ------
| BC Break      | yes

### Description
Moved laminas-validator to the require section and bumped directly the version to `^3.0`. This required to update some more dependencies and fix issues in the factories afterward. 

I tried to make this PR as small as possible and didn't added typehints for methods that doesn't require it yet. 

Relates to #114

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -34,16 +34,16 @@
     "require": {
         "php": "~8.1.0 || ~8.2.0 || ~8.3.0 || ~8.4.0",
         "laminas/laminas-eventmanager": "^3.12",
-        "laminas/laminas-servicemanager": "^3.22",
-        "laminas/laminas-stdlib": "^3.18"
+        "laminas/laminas-servicemanager": "^4.0",
+        "laminas/laminas-stdlib": "^3.18",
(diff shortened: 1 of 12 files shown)


froschdesign:
@marcelthole 
You are adding support for laminas-servicemanager with version 4 here. Do you see a change to remove the dependency on the service manager?

marcelthole:
@froschdesign i guess it should be possible. The only real usage is the `\Laminas\Session\Service\ContainerAbstractServiceFactory` that could be removed.

Than we could document it in the Migration guide that the consumer must use something like this:
```php
'MySessionContainer' => static fn (ContainerInterface $c): \Laminas\Session\Container => new \Laminas\Session\Container('MySessionContainerNamespace', $c->get(ManagerInterface::class));
```

But i guess it make sense to split this in a second PR? Than i would create a Issue and PR for that one 👍

marcelthole:
Ahh i see there is already a duplicate in #103

arhimede:
this PR was implemented in #103

PR #127, opened 2025-04-25
Moved laminas-validator to require

SergiuBota1:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | no
| BC Break      | no

### Description

Addresses [Issue 114](https://github.com/laminas/laminas-session/issues/114) for version 2, as updating `laminas-validator` to version  ^3 would require BC breaks.

### DIFF
--- a/.laminas-ci/install-mongodb-extension-via-pecl.sh
+++ b/.laminas-ci/install-mongodb-extension-via-pecl.sh
@@ -10,6 +10,6 @@ fi
 set +e
 apt install make
 
-pecl install mongodb
+pecl install mongodb-1.21.0
 echo "extension=mongodb.so" > /etc/php/${PHP_VERSION}/mods-available/mongodb.ini
 phpenmod -v ${PHP} -s cli mongodb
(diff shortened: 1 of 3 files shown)


gsteel:
#114 is not relevant to the 2.x series. It is slated for 3.0 as denoted by the milestone, so BC breaks are expected.

arhimede:
@gsteel 
Take a look at this line:
https://github.com/laminas/laminas-session/blob/80a4108b3f451d0fb3afd1d63fc8a79ead4ddae6/src/Validator/Csrf.php#L8C1-L8C41
Validator is called directly.

this PR technically will fix a bug in version 2.x 
Should I open a new issue with that bug only ?

gsteel:
Historically, laminas have used "soft dependencies", i.e. we use symbols from libs that are not `require`d and then add those libs to `suggest`. I personally hate this, so fully endorse moving stuff to `require`, but, we tend to get complaints from users who, for example, don't use the validator component but do use session.

Seeing as you're dealing with this in 3.0.x, I don't think it's necessary to "fix" this in 2.x.

BTW, you can't change deps in a patch release, it'll have to be a minor _(Which this PR targets)_

The switch to v3 of validator is not a concern, because if users are on v2 of session, they _must_ have validator v2 required in order to use the CSRF validator, therefore, on upgrade to session v3, they will also need to upgrade validator too, i.e. the dependency changes will be obvious to them and they can't end up in a state where they have incompatible deps installed

PR #132, opened 2025-05-06
Removed laminas/laminas-http dependency

### LINKED ISSUE #130: Remove laminas-http dependency
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    | yes

#### Summary

laminas-http is used only to get the IP address .
And validate it against the one already saved in session. 

We can remove that depency and implement another way to get the IP address

This is the function call for laminas-http in question 
https://github.com/laminas/laminas-session/blob/6e797b8a609b62fd4c461805c96d49a3361dc78e/src/Validator/RemoteAddr.php#L118

SergiuBota1:
|    Q          |   A
|-------------- | ------
| BC Break      | yes
| New Feature   | yes
| RFC           | no

### Description

Removed laminas-http dependency and imported the necessary methods from the  [laminas-http](https://github.com/laminas/laminas-http/blob/2.23.x/src/PhpEnvironment/RemoteAddress.php). Let me know if tests should be imported as well.


froschdesign:
@SergiuBota1 
As this is a new major version, the validator can and must also be reworked. Simply copying it is not a solution. And, of course, tests must be available and complete.
Some help to rework a legacy validator: https://docs.laminas.dev/laminas-validator/v3/migration/refactoring-legacy-validators/ 

Thanks in advance! 👍🏻

marcelthole:
relates to issue #130 (would prevent that someone else only looks at the issues and not the open PRs and also works on it, like me in the past 😅 )

SergiuBota1:
> @SergiuBota1 As this is a new major version, the validator can and must also be reworked. Simply copying it is not a solution. And, of course, tests must be available and complete. Some help to rework a legacy validator: https://docs.laminas.dev/laminas-validator/v3/migration/refactoring-legacy-validators/
> 
> Thanks in advance! 👍🏻

Should i keep implementing `Laminas\Session\Validator\ValidatorInterface` or extend `AbstractValidator` from `laminas-validator`?

arhimede:
We have a question here:

Should we keep implementing `Laminas\Session\Validator\ValidatorInterface` or extend `AbstractValidator` from laminas-validator?

The question is valid for all validators from the `laminas-session`  package .

CSRF validator is extending anyway the `AbstractValidator` from `laminas-validator `, so we have here a hard dependency. 
https://github.com/laminas/laminas-session/blob/a9b613348b3770f31ce2fffea949baeb11b059fa/src/Validator/Csrf.php#L37

froschdesign:
If error messages need to be output, a translation is required and this is provided by the abstract class of laminas-validator.

src/Validator/RemoteAddr.php
```diff
@@ -31,33 +34,33 @@ class RemoteAddr implements SessionValidator
...
-     * @param null|string $data
+     * @param OptionsArgument $options
      */
-    public function __construct($data = null)
+    public function __construct(?string $data = null, array $options = [])
     {
+        $proxyHeader = $options['proxy_header'] ?? 'X_FORWARDED_FOR';
+
+        $this->useProxy       = isset($options['use_proxy']) && (bool) $options['use_proxy'];
+        $this->trustedProxies = isset($options['trusted_proxies']) ? (array) $options['trusted_proxies'] : [];
```
Ocramius: Given `OptionsArgument`, I wouldn't perform any casts here anymore

test/Validator/RemoteAddrTest.php
```diff
@@ -6,14 +6,23 @@
...
  * @covers \Laminas\Session\Validator\RemoteAddr
  */
 class RemoteAddrTest extends TestCase
 {
-    /** @var array */
-    protected $backup;
+    protected array $backup;
+
+    protected RemoteAddr $defaultRemoteAddr;
+    private ReflectionObject $remoteAddrReflection;
```
Ocramius: You seem to only need `getIpAddress()` from the object: perhaps reflect that?

test/Validator/RemoteAddrTest.php
```diff
@@ -143,10 +164,102 @@ public function testCanSpecifyWhichHeaderToUseStatically(): void
...
+            'use_proxy'       => true,
+            'trusted_proxies' => [
+                '192.168.0.10',
+                '10.0.0.1',
+            ],
+        ];
+
+        $reflectionMethod   = $this->remoteAddrReflection->getMethod('getIpAddress');
+        $remoteAddr         = new RemoteAddr(null, $options);
+        $getClientIpAddress = (string) $reflectionMethod->invoke($remoteAddr);
```
Ocramius: Can this (and following) test be written without relying on reflection?

SergiuBota1: To test without reflection we should set `getIpAddress` public, I don't see the point in doing that considering it is only used internally.

Ocramius: Right, but can we skip testing `getIpAddress` entirely, and instead just verify the public API?

arhimede:
Can somebody take a look at this one ?
This PR is conflicting with the #145

arhimede:
thanks @SergiuBota1

PR #136, opened 2025-05-12
Modernized http-user-agent validator

### LINKED ISSUE #133: refactor validator: http-user-agent
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    | yes

#### Summary

refactor and modernise the validator related to HTTP USER AGENT

SergiuBota1:
|    Q          |   A
|-------------- | ------
| BC Break      | yes
| New Feature   | yes
| RFC           | no

### Description
Pull request based on #133


src/Validator/HttpUserAgent.php
```diff
@@ -1,23 +1,16 @@
...
-    protected $data;
-
     /**
      * Constructor
      * get the current user agent and store it in the session as 'valid data'
-     *
-     * @param string|null $data
      */
-    public function __construct($data = null)
+    public function __construct(protected ?string $data = null)
```
gsteel: IMO, super-globals i.e. `['HTTP_USER_AGENT']` here should not be referenced in this class and there is a design problem in all the validators.

Constructor accepts arbitrary string, defaulting to a super-global but `isValid` depends on the super-global matching.

A better design, assuming that all validators require access to the "Environment" would be:

```php
class SomeValidator implements SessionValidator {
    public function __construct(private EnvironmentValueObject $env) {}
    
    public function isValid(EnvironmentValueObject $data): bool {
        if ($this->env->httpUserAgent !== $data->httpUserAgent) {
            // The user agent changed mid session
            return false;
        }
        
        return true;
    }
}
```

So, I would introduce a value object to represent the environment variables we care about - this could be populated from server globals, PSR request, fake data for testing etc, and validate the initial value has not changed.

I may be barking up the wrong tree, but anytime I see super-globals hard-coded inside a class it sets off alarm bells for me, not to mention that this validator will just be comparing `null === null` if the `User-Agent` header is not being sent at all.

alexmerlin: In this case I see two options:
- populate the `EnvironmentValueObject` (from the `$request`) from outside the package and maybe require `SomeValidator` to inject the validator via constructor or by calling a setter
- store a pre-populated `EnvironmentValueObject` in the `$container` which can then be injected where needed

gsteel: If this theoretical `EnvironmentValueObject` is to be introduced, static named constructors could be added, for example, `EnvironmentValueObject::fromGlobals()`, `EnvironmentValueObject::fromPsr7Request()` etc could ease the situation.

As a major release, I don't think it's a big deal to ask consumers to inject this dependency wherever it makes most sense.

I think that putting it in the container (Assume you mean PSR-11 container) is a bad idea - this is per-request information and ideally, stuff in your container should be stateless services.

At the point of session initialisation, the request should be available somewhere, so the initial "Environment" can be serialised into the session for comparison, then whatever has the responsibility of running the validators should be able to fetch and hydrate this initial state from the session for injection into the validators and comparison with the current value.

I've got no concrete idea of how this will work in this lib because I don't use it, so sorry to put a spanner in the works.

If it's not feasible for this release and requires too much refactoring or the current architecture prohibits this, then maybe it's something to defer for another day.

Maybe others on the @laminas/technical-steering-committee have a better overview of what is/isn't feasible with this lib.

PR #137, opened 2025-05-12
Modernized Id validator

SergiuBota1:
|    Q          |   A
|-------------- | ------
| BC Break      | yes
| New Feature   | yes
| RFC           | no

### Description
Pull request based on #134

PR #135, opened 2025-05-07
Modern codebase update (Step 1/x)

marcelthole:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| BC Break      | yes

### Description

According to #113 i created a first PR with some changes.

In this PR i 
- enabled the stric_types cs-code fixer rule again and added all strict types
- removed the deprecated compatibility files which should not be required anymore for a new major release
- removed the deprecated PHP 7.1 functions and trigger an E_USER_DEPRECATED in PHP 8.4 for the `sid_length` and `sid_bits_per_character` https://wiki.php.net/rfc/deprecations_php_8_4#sessionsid_length_and_sessionsid_bits_per_character
- Removed the `SID` constant in the `sessionExists` method. This has the effect that some tests are not doing the same anymore and the session does not exists after session_close calls. I didn't found an other solution for that. If this behavior change i fine for you, i will also update that one in the changelog for the v3 release

- [x] Clarify change of `SID` constant in `sessionExists` method and document the change in the v2 to v3 migration guide

I see there is more things to do like add more native method and return types but didn't want to make this PR too big and will create a follow up one.

Issue #113, opened 2025-03-05
update the entire code base


arhimede:
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    | yes

#### Summary

Update the codebase to use the PHP minimum version features , annotations, etc.

arhimede:
this is implemented already in several merged PR's

PR #140, opened 2025-05-13
Add native types and template annotation for the ValidatorInterface

marcelthole:
|    Q          |   A
|-------------- | ------
| BC Break      | yes

### Description

Based on issue #113 and continuing the PR #135 - added native method and return types to the "ValidatorInterface" only, to keep it easier to review.

This should hopefully not affect the existing PRs #137, #136 and #132 which aims to make the validator itself better. This PR does not aim to solve the same problems, only to make the classes match the interface.

### DIFF
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -1319,9 +1319,4 @@
       <code><![CDATA[$backup]]></code>
     </MissingConstructor>
   </file>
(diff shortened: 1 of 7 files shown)


arhimede (approved):

arhimede:
thanks @marcelthole

PR #143, opened 2025-05-14
PHP8.4: disable ID validator

MatthiasKuehneEllerhold:
|    Q          |   A
|-------------- | ------
| Documentation |no
| Bugfix        | yes
| BC Break      | no
| New Feature   | no
| RFC           | no
| QA            | no

### Description

- Have a project with PHP 8.3 and session.sid_bits_per_character > 4
- Visit web page to create a valid session ID
- Upgrade PHP to 8.4
- Visit web page - blank - "Session validation failed"
- Set attach_default_validators to false
- Visit web page again - still blank - "Session validation failed"
- Delete all session files OR delete the session cookie on all clients (unrealistic on any scale, but doable in dev)
- Visit web page again - session is gone and web page works again.

The reason is, that PHP 8.4+ ALWAYS returns "4" for session.sid_bits_per_character. But it seems like PHP 8.4 happily accepts valid session with > 4. It will just generate NEW session ID with just 4 bits per char.
So PHP accepts the old session ID, but the validator does not and it will crash out.

### DIFF
--- a/src/Validator/Id.php
+++ b/src/Validator/Id.php
@@ -9,6 +9,8 @@ use function session_id;
 use function strrpos;
 use function substr;
 
+use const PHP_VERSION_ID;
+
 /**
  * session_id validator
  */
@@ -47,12 +49,6 @@ class Id implements ValidatorInterface
      */
     public function isValid()
     {
-        if (PHP_VERSION_ID >= 80400) {
-            // PHP 8.4 deprecated session.sid_bits_per_character and set it to "4" hard.
-            // Old (pre PHP 8.4) session IDs with a higher bitrate are still valid though.
-            return true;
-        }
-
         $id          = $this->id;
         $saveHandler = ini_get('session.save_handler');
         if ($saveHandler === 'cluster') { // Zend Server SC, validate only after last dash
@@ -62,9 +58,15 @@ class Id implements ValidatorInterface
             }
         }
 
-        // Get the session id bits per character INI setting, using 5 if unavailable
-        $hashBitsPerChar = ini_get('session.sid_bits_per_character');
-        $hashBitsPerChar = is_numeric($hashBitsPerChar) ? (int) $hashBitsPerChar : 5;
+        if (PHP_VERSION_ID >= 80400) {
+            // PHP 8.4 deprecated session.sid_bits_per_character and set it hard to "4".
+            // Old (pre PHP 8.4) session IDs with a higher bitrate are still valid though.
+            $hashBitsPerChar = 6;
+        } else {
+            // Get the session id bits per character INI setting, using 5 if unavailable
+            $hashBitsPerChar = ini_get('session.sid_bits_per_character');
+            $hashBitsPerChar = is_numeric($hashBitsPerChar) ? (int) $hashBitsPerChar : 5;
+        }
 
         $pattern = match ($hashBitsPerChar) {
             4 => '#^[0-9a-f]*$#',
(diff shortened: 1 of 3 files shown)


arhimede:
can you fix the tests ?

marcelthole:
I checked the error description here as well and this could be reproduced in PHP 8.3 only if someone has already an Session with a `SessionStorage` 

```php
require __DIR__ . '/vendor/autoload.php';


$config = new \Laminas\Session\Config\SessionConfig();
$config->setSidBitsPerCharacter(6);
$manager = new Laminas\Session\SessionManager($config, null, null, [], [
    'attach_default_validators' => true
]);
$session = new Laminas\Session\Container('session', $manager);

echo 'PHP Version: ' . phpversion() . PHP_EOL;
echo 'Session Counter: ' . ++$session->myCounter . PHP_EOL;
var_dump($_SESSION);
```

In this case i have a session with the `Validator\Id` already attached in my session. So as soon as i open the page it doesnt make a difference anymore to set `attach_default_validators` to false. The Validator is still in the `$_SESSION`

Now i could remove the `$config->setSidBitsPerCharacter(6);` (required for PHP 8.4) or set it to `4` and get the error, because the Validator is still active in the Session. 

Possible Solutions:
1. `'attach_default_validators' => false` and wait until all sessions are expired and all new sessions doesn't contain the IdValidator anymore (could be problematic for long running sessions)
2. Delete all session after deployment (not possible for the most of the users)
3. Update the IdValidator to allow both formats regardingless of the `sid_bits_per_character` setting. 

~~And than i would also ask the Question if the Validator is still required and if it could be removed on the 3.0.x branch?~~
Without any validation at all there is maybe the possibility to inject code or XSS, so i revoked my question.

arhimede:
I would say that you can try to implement the solution #3 
> Update the IdValidator to allow both formats regardingless of the sid_bits_per_character setting.

The solution 1 is too hackish :-)

MatthiasKuehneEllerhold:
Ive implemented solution 3 in the newest version of the PR. Anything I should change?

>  Without any validation at all there is maybe the possibility to inject code or XSS, so i revoked my question.

Can you elaborate please?

arhimede:
Can we have a different opinion here ?
As for me this PR looks OK-ish, it is quite an obscure PHP feature

arhimede:
> The reason is, that PHP 8.4+ ALWAYS returns "4" for session.sid_bits_per_character.

Nope.
The default value it is indeed 4 
But you can change it in php.ini 
`session.sid_bits_per_character=5;`

or dynamically to whatever value you want 
`ini_set('session.sid_bits_per_character', '4');`

MatthiasKuehneEllerhold:
This is deprecated and leads to a warning.

src/Validator/Id.php
```diff
@@ -56,9 +58,15 @@ public function isValid()
             }
         }
 
-        // Get the session id bits per character INI setting, using 5 if unavailable
-        $hashBitsPerChar = ini_get('session.sid_bits_per_character');
-        $hashBitsPerChar = is_numeric($hashBitsPerChar) ? (int) $hashBitsPerChar : 5;
+        if (PHP_VERSION_ID >= 80400) {
+            // PHP 8.4 deprecated session.sid_bits_per_character and set it hard to "4".
+            // Old (pre PHP 8.4) session IDs with a higher bitrate are still valid though.
+            $hashBitsPerChar = 6;
```
alexmerlin, line 64: @MatthiasKuehneEllerhold Is there any reason for not defaulting to 5 bits on PHP 8.4+ too ?

MatthiasKuehneEllerhold: Because the user may have set 6 in a previous PHP version and these old session ID would be invalid then.

arhimede:
thanks @MatthiasKuehneEllerhold

PR #144, opened 2025-05-14
Strict manager interface

### LINKED ISSUE #113: update the entire code base

marcelthole:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| BC Break      | yes

### Description

Based on issue #113 and continuing the PR #135 - added native method and return types to the "ManagerInterface" only, to keep it easier to review.

There was also a small change in the option parsing logic for the SessionManager:
- The `setId()` method now only accepts `string` type instead of the previous `int|string` union type
- Options like `send_expire_cookie`, `clear_storage` and `preserve_storage` could be set during construction and will be used in `destroy()` and `start()` method calls if the parameter will be passed as null.

### DIFF
--- a/docs/book/v3/migration/v2-to-v3.md
+++ b/docs/book/v3/migration/v2-to-v3.md
@@ -39,3 +39,15 @@ The implementation of the `SessionManager::sessionExists()` method has been simp
 
 > NOTE: **Logical change**
 > Due to this implementation change, `sessionExists()` will now return `false` after `session_close()` has been called, whereas in version 2 it would return `true`.
+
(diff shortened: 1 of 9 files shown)


arhimede:
@marcelthole  can you rebase to solve the conflicts ?
I am about to merge this PR too

arhimede (approved):
LGTM
Thanks @marcelthole

PR #145, opened 2025-05-15
Refactor user agent validator

SergiuBota1:
|    Q          |   A
|-------------- | ------
| BC Break      | yes
| New Feature   | yes
| RFC           | no

### Description
Pull request based on #133 
I closed the previous pull request #136 and opened this one to keep it clean (and a failed rebase). Based on what was discussed I replaced the use of the super-global variable with an object populated from outside.

Do we want the properties in the object to be mutable?
Based on @gsteel [comment](https://github.com/laminas/laminas-session/pull/136#discussion_r2084941253) should we no longer allow `$data` to be null?


src/Validator/EnvironmentValueObject.php
```diff
@@ -0,0 +1,27 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Laminas\Session\Validator;
+
+class EnvironmentValueObject
```
gsteel: I was being verbose when I called it `EnvironmentValueObject`, just `Environment` or `HttpEnvironment` would be a better name for the class.

It should be final and immutable - it's a value object.

The constructor should accept _all_ of the values you might need this for, i.e.:

```php
private function __construct(
    public readonly string|null $userAgent,
    public readonly string|null $remoteIp,
    public readonly string|null $somethingElse,
    // ... non-exhaustive list
) {}
```

You would then add a named constructor/factory method for the situations you care about:

```php

public static function fromGlobals(array $server): self
{
    $userAgent = (isset($server['HTTP_USER_AGENT']) && is_string($server['HTTP_USER_AGENT']))
        ? $server['HTTP_USER_AGENT']
        : null;

    return new self($userAgent, ...);
}

public static function fromPsrRequest(ServerRequestInterface $request): self {
    // ...
}

// Example, verbose method name
public static function fromSerialisedSessionFormat(whatever): self;
```

Elsewhere in the codebase…

```php
$environment = HttpEnvironment::fromGlobals($_SERVER);
```

There should be **no setters and no getters**

src/Validator/HttpUserAgent.php
```diff
@@ -5,27 +5,18 @@
...
-    protected $data;
-
     /**
      * Constructor
      * get the current user agent and store it in the session as 'valid data'
-     *
-     * @param string|null $data
      */
-    public function __construct($data = null)
+    public function __construct(protected EnvironmentValueObject $env, protected ?string $data = null)
```
gsteel:
```suggestion
    public function __construct(private readonly EnvironmentValueObject $initial)
```

The environment passed to the constructor should represent the values you wish to compare against, so, they will probably be retrieved from the session - i.e. it's the data set that was recorded when the session was first started.

I don't know how these validators are supposed to work - if they are constructed also with the to-be-validated data, then the constructor signature would be:

```php
public function __construct(
    private readonly Environment $initial,
    private readonly Environment $current,
)
```

_(Perhaps)_

Then `isValid` is simply `return $this->initial->userAgent === $this->current->userAgent`

gsteel: Also, why do the validators have a `getData` method? - if this is only used in tests, it should be removed, it has no purpose.

src/Validator/EnvironmentValueObject.php
```diff
@@ -0,0 +1,27 @@
...
+{
+    public function __construct(protected ?string $httpUserAgent = null)
+    {
+    }
+
+    public static function fromGlobals(): self
+    {
+        return new self($_SERVER['HTTP_USER_AGENT'] ?? null);
+    }
```
gsteel: Don't hide the access to the global inside this method - it should be a parameter.

Imagine testing this method. As a parameter, you can easily simulate any values and you are not polluting a super-global for all and any subsequent tests.

src/Validator/EnvironmentValueObject.php
```diff
@@ -0,0 +1,27 @@
...
+
+    public function getHttpUserAgent(): ?string
+    {
+        return $this->httpUserAgent;
+    }
+
+    public function setHttpUserAgent(?string $httpUserAgent): void
+    {
+        $this->httpUserAgent = $httpUserAgent;
+    }
```
gsteel: Getters and setters be gone. Use `public readonly` properties with solid types

gsteel (changes requested):

SergiuBota1:
Uploaded an initial commit removing the `getData()` method - it was used and in `ValidatorChain.php`. Removing will require updating the other validators as well.
I'm aware of the failing test but i have questions related to [this use case](https://github.com/laminas/laminas-session/pull/145/files#diff-c052712441803f7f790b8f646229bcfae9ece43efc14ca7b13cd452f9b4a2fdcL74)

Should the data still be attached to the storage metadata going forward?

src/Validator/HttpUserAgent.php
```diff
@@ -36,17 +20,7 @@ public function __construct($data = null)
...
-        return $userAgent === $this->getData();
-    }
-
-    /**
-     * Retrieve token for validating call
-     */
-    public function getData(): mixed
-    {
-        return $this->data;
+        return $this->initial->userAgent === $this->current->userAgent;
```
gsteel, line 26: Not knowing this library well, is a situtation where `null === null` acceptable here?

It stands to reason that if a user has enabled this validator, then a `null` user-agent string would be a validation issue… maybe?

gsteel: OK, I just noticed the test case that asserts a `null` user-agent is deemed valid.

As you probably know this library better, is that behaviour still correct?

test/Validator/IdTest.php
```diff
@@ -45,9 +46,10 @@ public function testIsValidPhp71(int $bitsPerCharacter, string $id, bool $isVali
 
     public function testConstructorSetId(): void
```
gsteel: Testing constructors have set a property is a silly idea (Not yours, I know) - tests like this should be dropped in later PRs. This is fine for now.

src/ValidatorChain.php
```diff
@@ -69,9 +69,8 @@ private function attachValidator($event, $callback, $priority)
             array_unshift($callback, $test);
         }
         if ($context instanceof ValidatorInterface) {
-            $data = $context->getData();
             $name = $context->getName();
-            $this->getStorage()->setMetadata('_VALID', [$name => $data]);
+            $this->getStorage()->setMetadata('_VALID', [$name => '']);
```
gsteel: The initial validator state has to be persisted somewhere. It looks like in the current implementation, each validator is serialised as `['ValidatorName' => 'Whatever $data is']`

My suggestion would be to re-architect this so that an `Environment` value object is added to the session metadata when it is first initialised. Then, prior to validation, the necessary validators are constructed with
```php
new WhateverValidator(
    Environment::fromSessionMetaData(), // Or whatever naming makes sense
    Environment::fromRequestOrGlobals(),
);
```

src/Validator/HttpUserAgent.php
```diff
@@ -36,17 +20,8 @@ public function __construct($data = null)
...
-    }
-
-    /**
-     * Retrieve token for validating call
-     */
-    public function getData(): mixed
-    {
-        return $this->data;
+        $env = Environment::fromGlobals($_SERVER);
+        return $env->userAgent === $this->data;
```
gsteel: Referencing `$_SERVER` inside this method defeats the point of having the environment object passed around.

The validators should be **given** the initial state to validate against _and_ the current state to validate, not **fetch** the current state out of thin air.

The truth is, I may have led you down the garden path with my comments about pulling `$_SERVER` out of all the validators - it's the right thing to do, but it may not be possible with the current architecture.

The session manager initialises the validator chain and executes validation I believe - this is where the "environment" should be retrieved and injected into the validators, but, there is no clear access to the current request as far as I can see.

Here is some pseudo code from the context of the session manager that hopefully illustrates what I was getting at:

```php
// The contents of the session:
$storage = $this->getSessionStorage();

if ($storage->has('The Initial State of the Environment')) {
   // we already have a record of the env from a previous request
   $initialEnvironment = Environment::fromSessionData($storage->get('The Data'));
} else {
   // This is the first request
   $initialEnvironment = Environment::fromGlobals($_SERVER); // Or the MVC request object, or whatever else I can this data from
   $storage->set('The initial state', serialiseOrSomething($initialEnvironment));
}

// Validation
$currentEnvironment = Environment::fromGlobals($_SERVER);

// Make the chain:
$chain = new ValidationChain();
foreach ($this->configuredValidators() as $className => $options) {
    // Validator interface will need to define a specific constructor signature or Psalm will hate this
    $chain->attach(
        new $className($initialEnvironment, $currentEnvironment, $options),
    );
}

if (! $chain->isValid()) {
   // blow up
}
```

Given something like the above

- the validators no longer need to fetch data from globals because all the data they need is given to them.
- testing validators is as simple as creating `Environment` value objects in test cases
- there is only one place where globals are used
- the `Environment` object is consistent, reliable and serialisable

gsteel (changes requested):

src/SessionManager.php
```diff
@@ -160,16 +166,38 @@ public function start($preserveStorage = false)
...
+                $initialEnvironment = Environment::fromGlobals($_SERVER);
+                $this->getStorage()->setMetadata('environment', serialize($initialEnvironment));
+            }
+
+            if ($data !== null) {
+                /** @var Environment $currentEnvironment */
+                $currentEnvironment = unserialize($data);
+            } else {
+                $currentEnvironment = Environment::fromGlobals($_SERVER);
+            }
```
gsteel: This block should just be replaced with `$currentEnvironment = Environment::fromGlobals($_SERVER);`

We are initialising new validators and we should only care what the environment is right now, not any kind of stale or already serialised data

src/Validator/Environment.php
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    public static function getServerOption(string $name, ?array $superglobal = null): mixed
+    {
+        if ($superglobal === null) {
+            $superglobal = $_SERVER;
+        }
+
+        return $superglobal[$name] ?? null;
+    }
```
gsteel: Why is this method here?

I cannot see any use case for this. This value object should be a wrapper around common environment related/request variables so that we have a predictable API for retrieving the data, therefore any/all data retrieval should be via public properties.

src/Validator/Environment.php
```diff
@@ -0,0 +1,48 @@
...
+            ? $server['HTTP_USER_AGENT']
+            : null;
+
+        $remoteAddr = isset($server['REMOTE_ADDR']) && is_string($server['REMOTE_ADDR'])
+            ? $server['REMOTE_ADDR']
+            : null;
+
+        if ($remoteAddr === null || (isset($options['use_proxy']) && $options['use_proxy'])) {
+            $remoteAddr = RemoteAddr::getIpAddress($options, $remoteAddr);
+        }
```
gsteel: Accessing static methods on a validator here is the wrong thing to do - you are making this value object dependent on the validator - `RemoteAddr` needs the `X-ForwardedFor`, maybe `X-ForwardedProto` or whatever - these should be made public properties and simply copied from the given array.

```suggestion
        $forwardedFor = isset($server['X_FORWARDED_FOR']) && is_string($server['X_FORWARDED_FOR'])
            ? $server['X_FORWARDED_FOR']
            : null;
```

The `$options` argument should also be dropped

src/ValidatorChain.php
```diff
@@ -6,21 +6,33 @@
...
+             * @var class-string<ValidatorInterface> $validator
+             */
             foreach ($validators as $validator => $data) {
-                $this->attachValidator('session.validate', [new $validator($data), 'isValid'], 1);
+                $currentEnvironment = $data instanceof Environment ? $data : Environment::fromGlobals($_SERVER);
+                $this->attachValidator(
+                    'session.validate',
+                    [new $validator(unserialize($environment), $currentEnvironment), 'isValid'],
+                    1
+                );
```
gsteel: Correct me if I am wrong, but it looks to me like all of this work is already done in `SessionManager`. Why repeat it?

src/ValidatorChain.php
```diff
@@ -69,11 +81,9 @@ private function attachValidator($event, $callback, $priority)
             array_unshift($callback, $test);
         }
         if ($context instanceof ValidatorInterface) {
```
gsteel, line 65: Because the validator chain is specific to this lib, why not change this method signature to `attachValidator(Event $event, ValidatorInterface $validator, int $priority)` for better type inference? _(Possibly better in another patch)_

src/ValidatorChain.php
```diff
@@ -69,11 +81,9 @@ private function attachValidator($event, $callback, $priority)
...
-            $data = $context->getData();
             $name = $context->getName();
-            $this->getStorage()->setMetadata('_VALID', [$name => $data]);
+            $this->getStorage()->setMetadata('_VALID', [$name => serialize($context->current)]);
```
gsteel: Can you check - does `setMetadata` merge or replace data? If it's a replace operation (which its name suggests) then there could only ever be 1 validator in a re-constituted chain.

I don't really understand why the validators are persisted to metadata at all, as they are constructed inside `SessionManager` from a list of configuration items right?

src/Validator/Id.php
```diff
@@ -64,15 +54,7 @@ public function isValid(): bool
...
-        return (bool) preg_match($pattern, $id);
-    }
-
-    /**
-     * Retrieve token for validating call (session_id)
-     */
-    public function getData(): ?string
-    {
-        return $this->id;
+        return (bool) preg_match($pattern, $this->initial->sessionId);
```
gsteel: Surely you'd want to validate the "current" session id?

gsteel (changes requested):

SergiuBota1:
Based on this comment i've change de flow to only allow attaching validators via config file, the benefit of this is allowing to drop the unused `$data` in the attach method of `ValidatorChain`. Also removed test cases using that function.

> I don't really understand why the validators are persisted to metadata at all, as they are constructed inside SessionManager from a list of configuration items right?

However the old version allowed users to manually attach validators in other places during the request, if i understand correctly this should no longer be the case, correct?

arhimede:
@alexmerlin  Can you too review this one  ? We are a bit behind with this package

alexmerlin:
@SergiuBota1 Let's keep things straightforward and allow adding validators only via the configs.
If one would like to add a new validator at runtime (less likely to be needed), they should create a new instance of the validator chain.

Jurj-Bogdan:
> @SergiuBota1 Let's keep things straightforward and allow adding validators only via the configs.

Unless I'm mistaken this seems to be the case in the current version of this PR
For now i've only updated the branch to match current 3.0.x

arhimede:
@gsteel I am about to merge this PR , we are quite late with this :-)

arhimede:
@SergiuBota1  LGTM

arhimede (approved):

PR #149, opened 2025-06-12
Removed laminas/laminas-stdlib dependency and added native types to Storage module

Jurj-Bogdan:
|    Q          |   A
|-------------- | ------
| BC Break      | yes

### Description

Initial commit to add native method and return types to the `src/Storage` module, based on issue #113. 

This PR might also be the place to further refactor the module as needed but I'm unsure how what the end goal of that refactoring might be.

### DIFF
--- a/src/AbstractContainer.php
+++ b/src/AbstractContainer.php
@@ -468,7 +468,7 @@ abstract class AbstractContainer extends ArrayObject
      * @param  array<TKey, TValue>|Traversable $input
      * @return array<TKey, TValue>
      */
-    public function exchangeArray(mixed $input): array
+    public function exchangeArray(object|array $input): array
     {
         $input = is_array($input) ? $input : iterator_to_array($input);
 
--- a/src/AbstractContainer.php
+++ b/src/AbstractContainer.php
@@ -7,6 +7,7 @@ namespace Laminas\Session;
 use AllowDynamicProperties;
 use ArrayIterator;
 use ArrayObject;
+use Iterator;
 use Laminas\Session\ManagerInterface as Manager;
 use Laminas\Session\Storage\StorageInterface as Storage;
 use Traversable;
@@ -496,16 +497,14 @@ abstract class AbstractContainer extends ArrayObject
     /**
      * Create a new iterator from an ArrayObject instance
      */
-    public function getIterator(): Traversable
+    public function getIterator(): Iterator
     {
         $this->expireKeys();
         $storage   = $this->getStorage();
         $container = $storage[$this->getName()];
-
-        if ($container instanceof Traversable) {
+        if ($container instanceof Iterator) {
             return $container;
         }
-
         return new ArrayIterator($container);
     }
 
--- a/src/AbstractContainer.php
+++ b/src/AbstractContainer.php
@@ -4,12 +4,11 @@ declare(strict_types=1);
 
 namespace Laminas\Session;
 
-use AllowDynamicProperties;
 use ArrayIterator;
 use ArrayObject;
-use Iterator;
 use Laminas\Session\ManagerInterface as Manager;
 use Laminas\Session\Storage\StorageInterface as Storage;
+use ReturnTypeWillChange;
 use Traversable;
 
 use function array_filter;
@@ -20,6 +19,7 @@ use function assert;
 use function is_array;
 use function is_scalar;
 use function is_string;
+use function iterator_to_array;
 use function preg_match;
 use function time;
 
@@ -35,7 +35,6 @@ use function time;
  * @template TValue
  * @template-extends ArrayObject<TKey, TValue>
  */
-#[AllowDynamicProperties]
 abstract class AbstractContainer extends ArrayObject
 {
     /**
@@ -89,7 +88,7 @@ abstract class AbstractContainer extends ArrayObject
     /**
      * Set the default ManagerInterface instance to use when none provided to constructor
      */
-    public static function setDefaultManager(?Manager $manager = null): void
+    public static function setDefaultManager(?Manager $manager): void
     {
         static::$defaultManager = $manager;
     }
@@ -165,7 +164,7 @@ abstract class AbstractContainer extends ArrayObject
     /**
      * Create a new container object on which to act
      */
-    protected function createContainer(): Traversable
+    protected function createContainer(): ArrayObject
     {
         return new ArrayObject([], ArrayObject::ARRAY_AS_PROPS);
     }
@@ -385,6 +384,7 @@ abstract class AbstractContainer extends ArrayObject
      * Set Offset
      *
      * @param TKey|non-empty-string $key
+     * @param TValue $value
      */
     public function __set(string $key, mixed $value): void
     {
@@ -465,47 +465,40 @@ abstract class AbstractContainer extends ArrayObject
     /**
      * Exchange the array for another one.
      *
-     * @param array<TKey, TValue>|object $input
+     * @param  array<TKey, TValue>|Traversable $input
      * @return array<TKey, TValue>
      */
     public function exchangeArray(mixed $input): array
     {
-        // handle arrayobject, iterators and the like:
-        if ($input instanceof ArrayObject) {
-            $input = $input->getArrayCopy();
-        }
-        if (! is_array($input)) {
-            $input = (array) $input;
-        }
+        $input = is_array($input) ? $input : iterator_to_array($input);
 
         $storage = $this->verifyNamespace();
         $name    = $this->getName();
 
+        /** @var array<string, mixed>|Traversable<string, mixed> $old */
         $old            = $storage[$name];
         $storage[$name] = $input;
-        if ($old instanceof ArrayObject) {
-            /** @var array<TKey, TValue> $array */
-            $array = $old->getArrayCopy();
-            return $array;
-        }
 
-        /** @var array<TKey, TValue> $array */
-        $array = $old;
-        return $array;
+        /** @var array<TKey, TValue> $return */
+        $return = is_array($old) ? $old : iterator_to_array($old);
+        return $return;
     }
 
     /**
      * Create a new iterator from an ArrayObject instance
+     *
+     * @return Traversable<TKey, TValue>
      */
-    public function getIterator(): Iterator
+    #[ReturnTypeWillChange]
+    public function getIterator(): Traversable
     {
         $this->expireKeys();
         $storage   = $this->getStorage();
         $container = $storage[$this->getName()];
-        if ($container instanceof Iterator) {
-            return $container;
-        }
-        return new ArrayIterator($container);
+
+        /** @var Traversable<TKey, TValue> $return */
+        $return = $container instanceof Traversable ? $container : new ArrayIterator($container);
+        return $return;
     }
 
     /**
@@ -591,14 +584,9 @@ abstract class AbstractContainer extends ArrayObject
         $storage   = $this->verifyNamespace();
         $container = $storage[$this->getName()];
 
-        if ($container instanceof ArrayObject) {
-            /** @var array<TKey, TValue> $array */
-            $array = $container->getArrayCopy();
-            return $array;
-        }
-
         /** @var array<TKey, TValue> $array */
-        $array = $container;
+        $array = $container instanceof ArrayObject ? $container->getArrayCopy() : $container;
+
         return $array;
     }
 }
--- a/src/AbstractContainer.php
+++ b/src/AbstractContainer.php
@@ -4,19 +4,21 @@ declare(strict_types=1);
 
 namespace Laminas\Session;
 
+use AllowDynamicProperties;
 use ArrayIterator;
+use ArrayObject;
 use Laminas\Session\ManagerInterface as Manager;
 use Laminas\Session\Storage\StorageInterface as Storage;
-use Laminas\Stdlib\ArrayObject;
 use Traversable;
 
 use function array_filter;
 use function array_flip;
 use function array_keys;
 use function array_map;
+use function assert;
 use function is_array;
-use function is_object;
 use function is_scalar;
+use function is_string;
 use function preg_match;
 use function time;
 
@@ -32,48 +34,41 @@ use function time;
  * @template TValue
  * @template-extends ArrayObject<TKey, TValue>
  */
+#[AllowDynamicProperties]
 abstract class AbstractContainer extends ArrayObject
 {
     /**
      * Container name
-     *
-     * @var string
      */
-    protected $name;
+    protected string $name;
 
-    /** @var Manager */
-    protected $manager;
+    protected Manager $manager;
 
     /**
      * Default manager class to use if no manager has been provided
-     *
-     * @var string
      */
-    protected static $managerDefaultClass = SessionManager::class;
+    protected static string $managerDefaultClass = SessionManager::class;
 
     /**
      * Default manager to use when instantiating a container without providing a ManagerInterface
-     *
-     * @var Manager
      */
-    protected static $defaultManager;
+    protected static ?Manager $defaultManager = null;
 
     /**
      * Default value to return by reference from offsetGet
      *
-     * @var mixed
+     * @phpcs:disable WebimpressCodingStandard.Classes.NoNullValues.Invalid
      */
-    private $defaultValue;
+    private mixed $defaultValue = null;
 
     /**
      * Constructor
      *
      * Provide a name ('Default' if none provided) and a ManagerInterface instance.
-     *
-     * @param  null|string                        $name
+
      * @throws Exception\InvalidArgumentException
      */
-    public function __construct($name = 'Default', ?Manager $manager = null)
+    public function __construct(string $name = 'Default', ?Manager $manager = null)
     {
         if (! preg_match('/^[a-z0-9][a-z0-9_\\\\]+$/i', $name)) {
             throw new Exception\InvalidArgumentException(
@@ -92,10 +87,8 @@ abstract class AbstractContainer extends ArrayObject
 
     /**
      * Set the default ManagerInterface instance to use when none provided to constructor
-     *
-     * @return void
      */
-    public static function setDefaultManager(?Manager $manager = null)
+    public static function setDefaultManager(?Manager $manager = null): void
     {
         static::$defaultManager = $manager;
     }
@@ -105,10 +98,9 @@ abstract class AbstractContainer extends ArrayObject
      *
      * If none provided, instantiates one of type {@link $managerDefaultClass}
      *
-     * @return Manager
      * @throws Exception\InvalidArgumentException If invalid manager default class provided.
      */
-    public static function getDefaultManager()
+    public static function getDefaultManager(): Manager
     {
         if (null === static::$defaultManager) {
             $manager = new static::$managerDefaultClass();
@@ -125,10 +117,8 @@ abstract class AbstractContainer extends ArrayObject
 
     /**
      * Get container name
-     *
-     * @return string
      */
-    public function getName()
+    public function getName(): string
     {
         return $this->name;
     }
@@ -136,10 +126,9 @@ abstract class AbstractContainer extends ArrayObject
     /**
      * Set session manager
      *
-     * @return Container
      * @throws Exception\InvalidArgumentException
      */
-    protected function setManager(?Manager $manager = null)
+    protected function setManager(?Manager $manager = null): static
     {
         if (null === $manager) {
             $manager = static::getDefaultManager();
@@ -156,10 +145,8 @@ abstract class AbstractContainer extends ArrayObject
 
     /**
      * Get manager instance
-     *
-     * @return Manager
      */
-    public function getManager()
+    public function getManager(): Manager
     {
         return $this->manager;
     }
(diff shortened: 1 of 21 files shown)


Jurj-Bogdan:
Speaking of refactoring, I'd guess that `SessionStorage` and `SessionArrayStorage` could be marked as final as a start

src/Storage/AbstractSessionArrayStorage.php
```diff
@@ -40,24 +41,16 @@ abstract class AbstractSessionArrayStorage implements
...
     }
 
     /**
      * Initialize Storage
-     *
-     * @param  array $input
-     * @return void
      */
-    public function init($input = null)
+    public function init(array|ArrayObject|null $input = null): void
```
samsonasik: on input check, `is_object` is too generic imo, if that's going to be more specific, using `?iterable` probably better, then use `instanceof Traversable` for `iterator_to_array()` usage.

src/Storage/AbstractSessionArrayStorage.php
```diff
@@ -74,39 +67,39 @@ public function init($input = null)
     /**
      * Get Offset
      *
-     * @return mixed
+     * @param non-empty-string $key
      */
-    public function __get(mixed $key)
+    public function __get(mixed $key): mixed
```
gsteel:
```suggestion
    public function __get(string $key): mixed
```

src/Storage/AbstractSessionArrayStorage.php
```diff
@@ -74,39 +67,39 @@ public function init($input = null)
...
     }
 
     /**
      * Set Offset
      *
-     * @return void
+     * @param non-empty-string $key
      */
-    public function __set(mixed $key, mixed $value)
+    public function __set(mixed $key, mixed $value): void
```
gsteel:
```suggestion
    public function __set(string $key, mixed $value): void
```

src/Storage/AbstractSessionArrayStorage.php
```diff
@@ -74,39 +67,39 @@ public function init($input = null)
...
     }
 
     /**
      * Isset Offset
      *
-     * @return bool
+     * @param non-empty-string $key
      */
-    public function __isset(mixed $key)
+    public function __isset(mixed $key): bool
```
gsteel:
```suggestion
    public function __isset(string $key): bool
```

src/Storage/AbstractSessionArrayStorage.php
```diff
@@ -74,39 +67,39 @@ public function init($input = null)
...
     }
 
     /**
      * Unset Offset
      *
-     * @return void
+     * @param non-empty-string $key
      */
-    public function __unset(mixed $key)
+    public function __unset(mixed $key): void
```
gsteel:
```suggestion
    public function __unset(string $key): void
```

src/Storage/AbstractSessionArrayStorage.php
```diff
@@ -123,82 +116,77 @@ public function __destruct()
...
         return isset($_SESSION[$key]);
     }
 
     /**
      * Offset Get
      *
-     * @return mixed
+     * @param non-empty-string $key
      */
     #[ReturnTypeWillChange]
```
gsteel: All usage of `#[ReturnTypeWillChange]` should be removed once types are added

src/Storage/AbstractSessionArrayStorage.php
```diff
@@ -123,82 +116,77 @@ public function __destruct()
...
-    public function unserialize($session)
+    public function unserialize(string $session): mixed
     {
         return unserialize($session);
     }
 
     /** @inheritDoc */
     #[ReturnTypeWillChange]
-    public function getIterator()
+    public function getIterator(): ArrayIterator
```
gsteel: There's no need to narrow this type to the exact iterator here

```suggestion
    public function getIterator(): Traversable
```

src/Storage/AbstractSessionArrayStorage.php
```diff
@@ -207,10 +195,8 @@ public function getIterator()
      * Load session object from an existing array
      *
      * Ensures $_SESSION is set to an instance of the object when complete.
-     *
-     * @return SessionStorage
      */
-    public function fromArray(array $array)
+    public function fromArray(array $array): self
```
gsteel:
```suggestion
    public function fromArray(array $array): static
```

This should probably be changed in the interface too. Returning `static` preserves your type, i.e. `$type = SomeConcrete::fromArray(['foo'])` will mean `$type instanceof SomeConcrete` rather than `SomeInterface`

src/Storage/AbstractSessionArrayStorage.php
```diff
@@ -221,10 +207,8 @@ public function fromArray(array $array)
 
     /**
      * Mark object as isImmutable
-     *
-     * @return SessionStorage
      */
-    public function markImmutable()
+    public function markImmutable(): self
```
gsteel: Again, maybe `static` here

src/Storage/AbstractSessionArrayStorage.php
```diff
@@ -233,21 +217,16 @@ public function markImmutable()
...
     }
 
     /**
      * Lock this storage instance, or a key within it
-     *
-     * @param  null|int|string $key
-     * @return $this
      */
-    public function lock($key = null)
+    public function lock(null|int|string $key = null): self
```
gsteel: Are we sure that `int` is relevant here? AFAIK, top-level session keys must be strings.

You can write `$_SESSION[123] = 'foo';` without PHP complaining, but the key `123` will not get persisted

src/Storage/AbstractSessionArrayStorage.php
```diff
@@ -263,11 +242,8 @@ public function lock($key = null)
 
     /**
      * Is the object or key marked as locked?
-     *
-     * @param  null|int|string $key
-     * @return bool
      */
-    public function isLocked($key = null)
+    public function isLocked(null|int|string $key = null): bool
```
gsteel: Same thing about `int` here too

src/Storage/AbstractSessionArrayStorage.php
```diff
@@ -300,11 +276,8 @@ public function isLocked($key = null)
 
     /**
      * Unlock an object or key marked as locked
-     *
-     * @param  null|int|string $key
-     * @return $this
      */
-    public function unlock($key = null)
+    public function unlock(null|int|string $key = null): self
```
gsteel: `int`?

src/Storage/AbstractSessionArrayStorage.php
```diff
@@ -409,11 +377,9 @@ public function getMetadata($key = null)
     /**
      * Clear the storage object or a subkey of the object
      *
-     * @param  null|int|string            $key
-     * @return $this
      * @throws Exception\RuntimeException
      */
-    public function clear($key = null)
+    public function clear(null|int|string $key = null): self
```
gsteel: `int`?

src/Storage/ArrayStorage.php
```diff
@@ -350,13 +315,19 @@ public function fromArray(array $array)
...
+    public function toArray(bool $metadata = false): array
     {
+        /** @var iterable $values */
         $values = $this->getArrayCopy();
-        if ($metaData) {
+
+        if (is_object($values)) {
+            $values = ArrayUtils::iteratorToArray($values);
+        }
+
```
gsteel: If we can keep `Laminas\Stdlib` out of this library, it's one less dependency to worry about

`$values = is_array($values) ? $values : iterator_to_array($values);`

src/Storage/Factory.php
```diff
@@ -28,28 +28,29 @@ abstract class Factory
     /**
      * Create and return a StorageInterface instance
      *
-     * @param  string                             $type
-     * @param  array|Traversable                  $options
-     * @return StorageInterface
      * @throws Exception\InvalidArgumentException For unrecognized $type or individual options.
      */
-    public static function factory($type, $options = [])
+    public static function factory(?string $type, iterable $options = []): StorageInterface
```
gsteel:
```suggestion
    public static function factory(string $type, array $options = []): StorageInterface
```

Make this not nullable and drop the exception below.

Also, It's probably unnecessary for `$options` to be `iterable` - Hang over from the days of `Laminas\Config` IMO. Requiring an array means we don't have to use `ArrayUtils`

src/Storage/SessionArrayStorage.php
```diff
@@ -16,21 +16,19 @@ class SessionArrayStorage extends AbstractSessionArrayStorage
     /**
      * Get Offset
      *
-     * @param  mixed $key
-     * @return mixed
+     * @param non-empty-string $key
      */
-    public function &__get($key)
+    public function &__get(mixed $key): mixed
```
gsteel:
```suggestion
    public function &__get(string $key): mixed
```

src/Storage/SessionStorage.php
```diff
@@ -26,15 +26,11 @@ class SessionStorage extends ArrayStorage
      *
      * Sets the $_SESSION superglobal to an ArrayObject, maintaining previous
      * values if any discovered.
-     *
-     * @param array|null $input
-     * @param int        $flags
-     * @param string     $iteratorClass
```
gsteel, line 32: `@param class-string<Iterator> $iteratorClass` ??

gsteel (changes requested):
I'm not convinced that it's possible for session keys to be integers, so the key type could be refined to `string` across the board. You can also type the interface to `@extends ArrayAccess<string, mixed>` etc.

It should be possible to eliminate usage of `Laminas\Stdlib`.

It's unfortunate that this implements `ArrayAccess` so we're stuck with `offsetSet(mixed $key, mixed $value): void` etc - `ArrayAccess` methods could have assertions for string keys perhaps because the doc blocks will not be inferred via SA tools AFAIK

Jurj-Bogdan:
Since the storage module was the main place `laminas/laminas-stdlib` was used (only 1 `ArrayUtils` use case in `SessionManager` outside of it), i've removed the dependency completely as per @gsteel's suggestion, replacing `Stdlib\ArrayObject`s with the native version.

src/AbstractContainer.php
```diff
@@ -168,20 +156,16 @@ public function getManager()
...
         return $this->getManager()->getStorage();
     }
 
     /**
      * Create a new container object on which to act
-     *
-     * @return ArrayObject
      */
-    protected function createContainer()
+    protected function createContainer(): Traversable
```
gsteel: Inheritors will likely need more than just `Traversable` for a container because I assume they will be writing to whatever this method returns, so, at the least `ArrayAccess` perhaps?

I'm not sure how this method is used. Can it be deleted entirely so that "Container" implementations just create whatever object they want to work with?

Alternatively, retain `ArrayObject`

src/AbstractContainer.php
```diff
@@ -463,11 +462,16 @@ public function offsetUnset($offset)
...
+     *
+     * @param array<TKey, TValue>|object $input
+     * @return array<TKey, TValue>
+     */
+    public function exchangeArray(mixed $input): array
     {
         // handle arrayobject, iterators and the like:
-        if (is_object($input) && ($input instanceof ArrayObject || $input instanceof \ArrayObject)) {
+        if ($input instanceof ArrayObject) {
             $input = $input->getArrayCopy();
```
gsteel: The input here is expected to be an array, so I suggest we blow up if `$input` is not. Psalm will call docblock contradiction which can be silenced.

If it's necessary to support `Traversable` we should probably `iterator_to_array` the input rather than testing specifically for an `ArrayObject`?

src/AbstractContainer.php
```diff
@@ -480,48 +484,47 @@ public function exchangeArray($input)
...
+        $array = $old;
+        return $array;
     }
 
-    /** @inheritDoc */
-    public function getIterator()
+    /**
+     * Create a new iterator from an ArrayObject instance
+     */
+    public function getIterator(): Iterator
```
gsteel: Why narrow the type to `Iterator`? what if `$container` is `IteratorAggregate`? `Traversable<TKey, TValue>` also aligns with `ArrayObject::getIterator()`

src/AbstractContainer.php
```diff
@@ -596,11 +586,19 @@ public function setExpirationHops($hops, $vars = null)
...
-        return $container instanceof ArrayObject ? $container->getArrayCopy() : $container;
+        if ($container instanceof ArrayObject) {
+            /** @var array<TKey, TValue> $array */
+            $array = $container->getArrayCopy();
+            return $array;
+        }
+
+        /** @var array<TKey, TValue> $array */
+        $array = $container;
+        return $array;
```
gsteel: This would be more readable with a temporary var before forcing the type:

```suggestion
        /** @var array<TKey, TValue> $array */
        $array = $container instanceof ArrayObject ? $container->getArrayCopy() : $container;
        
        return $array;
```

Loads of issues would be solved here if there were better types attributed to whatever `$this->getStorage()` returns. By the look of it, it's typically `StorageInterface<string, array<string, mixed>|ArrayObject<string, mixed>>`. (??)

src/AbstractContainer.php
```diff
@@ -32,48 +35,41 @@
  * @template TValue
  * @template-extends ArrayObject<TKey, TValue>
  */
+#[AllowDynamicProperties]
```
gsteel: Can I assume that the change in inheritance makes this required? i.e `Laminas\Stdlib\ArrayObject` already had this present?

src/Storage/AbstractSessionArrayStorage.php
```diff
@@ -28,41 +31,33 @@
...
  *
- * @see ReturnTypeWillChange
- *
- * @template TKey of array-key
+ * @template TKey of string
  * @template TValue
  * @template-implements IteratorAggregate<TKey, TValue>
  * @template-implements StorageInterface<TKey, TValue>
  */
+#[AllowDynamicProperties]
```
gsteel: Why is this now needed here?

src/Storage/AbstractSessionArrayStorage.php
```diff
@@ -28,41 +31,33 @@
...
-    public function init($input = null)
+    public function init(?iterable $input = null): void
     {
         if ((null === $input) && isset($_SESSION)) {
             $input = $_SESSION;
-            if (is_object($input) && ! $_SESSION instanceof ArrayObject) {
-                $input = (array) $input;
+            /** @var iterable $input */
+            if ($input instanceof Traversable && ! $_SESSION instanceof ArrayObject) {
+                $input = iterator_to_array($input);
```
gsteel: How about:

```php
$input ??= ($_SESSION ?? []);
$input = $input instanceof Traversable ? iterator_to_array($input) : $input;
$_SESSION = $input;
```

Those nested if's and yodas are horrible to read

src/Storage/AbstractSessionArrayStorage.php
```diff
@@ -123,94 +120,91 @@ public function __destruct()
...
+    /**
+     * Retrieve an external iterator
+     */
+    public function getIterator(): Traversable
     {
-        return new ArrayIterator($_SESSION);
+        /** @var Traversable<TKey, TValue> $return */
+        $return = new ArrayIterator($_SESSION);
+
+        return $return;
```
gsteel: Surprised forcing the var here is necessary… `$_SESSION` should already be `array<string, mixed>`.

Please also add `@return Traversable<TKey, TValue>`

src/Storage/AbstractSessionArrayStorage.php
```diff
@@ -123,94 +120,91 @@ public function __destruct()
...
 
     /**
      * Load session object from an existing array
      *
      * Ensures $_SESSION is set to an instance of the object when complete.
-     *
-     * @return SessionStorage
      */
-    public function fromArray(array $array)
+    public function fromArray(array $array): static
```
gsteel, line 207:
```suggestion
     * @param array<TKeyIn, TValIn> $array
     * @return static<TKeyIn, TValIn>
     *
     * @template TKeyIn of string
     * @template TValIn
     */
    public function fromArray(array $array): static
```

src/Storage/ArrayStorage.php
```diff
@@ -23,48 +25,40 @@
...
  *
- * @see ReturnTypeWillChange
- *
- * @template TKey of array-key
+ * @template TKey of string
  * @template TValue
  * @template-extends ArrayObject<TKey, TValue>
  * @template-implements StorageInterface<TKey, TValue>
  */
+#[AllowDynamicProperties]
```
gsteel: `#[AllowDynamicProperties]` is not required when `__get` and `__set` are defined:

https://3v4l.org/LcAVn

https://3v4l.org/8BGLr

src/Storage/StorageInitializationInterface.php
```diff
@@ -14,9 +14,6 @@ interface StorageInitializationInterface
 {
     /**
      * Initialize Session Storage
-     *
-     * @param  array $input
-     * @return void
      */
-    public function init($input = null);
+    public function init(?iterable $input = null): void;
```
gsteel, line 20: This could do with better type docs, i.e. `@param iterable<string, mixed>|null $input`

src/AbstractContainer.php
```diff
@@ -92,10 +88,8 @@ public function __construct($name = 'Default', ?Manager $manager = null)
 
     /**
      * Set the default ManagerInterface instance to use when none provided to constructor
-     *
-     * @return void
      */
-    public static function setDefaultManager(?Manager $manager = null)
+    public static function setDefaultManager(?Manager $manager = null): void
```
gsteel: I'd remove the `null` default assignment here so that users _must_ provide an argument, i.e. `Container::setDefaultManager(null)` to make intentions explicit

test/AbstractContainerTest.php
```diff
@@ -44,13 +34,10 @@ protected function setUp(): void
...
         $_SESSION = [];
-        Container::setDefaultManager(null);
+        Container::setDefaultManager();
     }
 
-    /**
-     * This test case fails on laminas-session 2.8.0 with the php error below and works fine on 2.7.*.
-     * "Only variable references should be returned by reference"
-     */
+    #[IgnoreDeprecations]
```
gsteel, line 40: I assume this is because of `__get` and `offsetGet` implementations returning by ref.

Perhaps, in a different patch, return by ref should be removed across the board and all these `IgnoreDeprecations` attributes removed.

gsteel (changes requested):
Thanks @Jurj-Bogdan - Pretty sure that most additions of `#[AllowDynamicProps]` can be removed if all implementations define `__set` etc.

src/AbstractContainer.php
Jurj-Bogdan: ArrayObject::getIterator() returns `Iterator` which while extending `Traversable` [triggered errors in a previous commit](https://github.com/laminas/laminas-session/actions/runs/16191740468/job/45708892458). I made a hotfix to match the signature to `ArrayObject`, but went back to returning `Traversable` now (albeit with the `#[\ReturnTypeWillChange]` attribute to suppress the error)

src/Storage/AbstractSessionArrayStorage.php
Jurj-Bogdan: I'm not a fan of this type "hack" either, but I went with them here and in some other cases to appease psalm (throwing `InvalidReturnType` and `InvalidReturnStatement` errors without it). Am i missing some other way of resolving them, or should i simply add them to the baseline?

src/AbstractContainer.php
Jurj-Bogdan: It seems to me that it's only used internally to generate a new container (or regenerate it if expiring by tts/hops) via `verifyNamespace()`, `expireByExpiryTime()` and `expireByHops()`. Would it be useful to extend it's functionality to allow users to change what object type to use?

gsteel: I dont think so - it's just a bucket for keys and values

src/AbstractContainer.php
gsteel: Ah, OK. Didn't realise it was narrowed in `ArrayObject` 👍

src/Storage/AbstractSessionArrayStorage.php
```diff
@@ -343,13 +330,12 @@ public function unlock($key = null)
      * - localizing session storage
      * - etc.
      *
-     * @param  string                     $key
-     * @param  mixed                      $value
-     * @param  bool                       $overwriteArray Whether to overwrite or merge array values; by default, merges
-     * @return $this
+     * $overwriteArray Whether to overwrite or merge array values; by default, merges
```
gsteel: Mising `@param bool ` here

gsteel (changes requested):

src/Storage/SessionArrayStorage.php
gsteel: @Jurj-Bogdan - Any reason why `$key` should still be mixed?

Jurj-Bogdan: My bad, must have managed to miss it among the changes.

src/AbstractContainer.php
```diff
@@ -468,7 +468,7 @@ public function offsetUnset(mixed $offset): void
      * @param  array<TKey, TValue>|Traversable $input
      * @return array<TKey, TValue>
      */
-    public function exchangeArray(mixed $input): array
+    public function exchangeArray(object|array $input): array
```
gsteel: If this is `mixed` in `\ArrayObject`, it can stay as mixed - This will reduce psalm issues.

Also, the parameter doc block should be updated to `array<TKey, TValue>|Traversable<TKey, TValue>`

Jurj-Bogdan: Quite a weird interaction as `ArrayObject` does have it as `object|array`, but psalm picks it up as `mixed`.

gsteel (approved):
Thanks for sticking with me @Jurj-Bogdan - We got there in the end 😅

PR #151, opened 2025-09-03
added native types to remaining classes missing them

Jurj-Bogdan:
|    Q          |   A
|-------------- | ------
| BC Break      | yes

### Description

Based on #113 - these classes should be the last ones needing native types added. 

While updating `Module.php` however, it seems to me that this class has no real use at the moment.
Unless i'm mistaken, this class was supposed to be used in `laminas-mvc` applications, and since it's now abandoned I believe `Module.php` can now be deleted altogether, right?

### DIFF
--- a/src/ValidatorChain.php
+++ b/src/ValidatorChain.php
@@ -33,10 +33,8 @@ class ValidatorChain extends EventManager
 
     /**
      * Retrieve session storage object
-     *
-     * @return StorageInterface
      */
-    public function getStorage()
+    public function getStorage(): StorageInterface
     {
         return $this->storage;
     }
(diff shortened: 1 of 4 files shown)


src/ValidatorChain.php
```diff
@@ -33,24 +33,17 @@ public function attach($eventName, callable $listener, $priority = 1)
...
     }
 
     /**
      * Internal implementation for attaching a listener to the
      * session validator chain.
-     *
-     * @param string   $event
-     * @param callable $callback
-     * @param int      $priority
-     * @return callable
```
gsteel, line 51: Can we accept and return a more precise type for `callable` - I don't know what the signature should be but, something like `callable(mixed): bool` perhaps?

gsteel: Actually, nevermind - I can see this extends from `EventManager` so types should be improved there. They are also plain `callable` in `EventManager` so we'll just end up with `MoreSpecificImplementedType` issues…

gsteel (approved):
Thanks @Jurj-Bogdan

gsteel:
I think we can leave `Module.php` where it is. You are right that it's probably useless, but it does mean interop for future forks etc - You never know, someone might step in and save MVC _(Unlikely though)_

PR #156, opened 2025-09-10
updated Id validator for php84 use

### LINKED ISSUE #153: Update Id validator in version 3.0.x for PHP 8.4
### Bug Report

|    Q        |   A
|------------ | ------
| Version(s)  | 3.0.x

#### Summary

Version 3.0.x of the `Id` validator will still raise the issue described in PR #143.
This PR should probably be implemented for the 3.0.x version as well, instead of the current deprecation warning only.

Jurj-Bogdan:
|    Q          |   A
|-------------- | ------
| Bugfix        | yes

### Description

Implemented @MatthiasKuehneEllerhold's PR #143 in branch 3.0.x, instead of the current deprecation notice.

Closes #153

### DIFF
--- a/src/Validator/Id.php
+++ b/src/Validator/Id.php
@@ -7,9 +7,7 @@ namespace Laminas\Session\Validator;
 use function ini_get;
 use function is_numeric;
 use function preg_match;
-use function trigger_error;
 
-use const E_USER_DEPRECATED;
 use const PHP_VERSION_ID;
 
 /**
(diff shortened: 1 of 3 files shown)


arhimede:
this is a bugfix from v2.25 which was fixed there, and this PR is porting it to V3 
@laminas/technical-steering-committee can we have a review here ?

src/Validator/Id.php
```diff
@@ -42,13 +40,15 @@ public function isValid(): bool
         }
 
         if (PHP_VERSION_ID >= 80400) {
-            trigger_error('session.sid_bits_per_character is deprecated starting with PHP 8.4', E_USER_DEPRECATED);
+            // PHP 8.4 deprecated session.sid_bits_per_character and set it hard to "4".
+            // Old (pre PHP 8.4) session IDs with a higher bitrate are still valid though.
+            $hashBitsPerChar = 6;
```
Ocramius, line 45: with this patch, the ID validation changes when upgrading `laminas/laminas-session`, causing session validation to fail.

Perhaps setting `4` (default by PHP) or `5` (our fallback) is healthier? :thinking: 

What value would we need to set here, to avoid a global production logout for users? Or is that an acceptable scenario, given the major version upgrade?

Ocramius (changes requested):

src/Validator/Id.php
MatthiasKuehneEllerhold: Its not a global logout... its a white page (500) for every user. 
You have to delete all sessions from your storage (either filesystem, redis or ...) in order for your user to be able to access the site again and be able to login again.
Thats what happened to our 8.4 upgrade and I guess the same will happen here if you set it to any other than 6.

Ocramius: The comment states that PHP 8.4 set it to `4` though? :thinking:

MatthiasKuehneEllerhold: Yes, but old sessions with 5 or 6 are still valid and in use by PHP. We had 5 in PHP 8.3, upgraded to 8.4 and all old sessions were still valid and in use by PHP, only this validator refused them and bricked the web page.

arhimede:
I am going to merge this 
thanks @Jurj-Bogdan

Ocramius:
Sorry, completely missed to check this again: thanks @arhimede

PR #157, opened 2025-09-12
Allow custom environment objects

### LINKED ISSUE #152: Adding custom validators restricted by the new `ValidatorInterface`
### BC Break Report

|    Q        |   A
|------------ | ------
| Version     | 3.x

#### Summary
Following the change to [the ValidatorInterface](https://github.com/laminas/laminas-session/blob/51d3179353a11a1f507284aafef981e0fd2525a5/src/Validator/ValidatorInterface.php#L15) ,  adding custom validators implementing it is restricted by the existing properties of the new [Environment](https://github.com/laminas/laminas-session/blob/3.0.x/src/Validator/Environment.php#L10) class.

Being final, the Environment can't be extended, so any user that would require using any other value from the `$_SERVER` will not be able to access it from the class, even if the custom validator will be required to pass the two instances of it.

Jurj-Bogdan:
|    Q          |   A
|-------------- | ------
| Documentation | no
| Bugfix        | yes
| RFC           | yes

### Description

Based on issue #152 , added a new `EnvironmentInterface` to allow users to create custom environments with additional properties if needed, and use them in current `SessionManager` flow.

Kept the existing `Environment` as the default for `SessionManager`, but users can provide a custom one via config under a new `environment` key.

```php
    'session_manager'    => [
        'environment' => AnyCustomEnvironment::class
    ],
```

Since properties can't be defined in the older supported  PHP versions, I've added getters to the `EnvironmentInterface` to force users to define the required properties for all existing validators, with any custom ones being extra.

One issue is that getters for public properties are unnecessary for the current `Environment` object, but psalm will throw a lot of `NoInterfaceProperties` in the validator classes, as the `Environment` parameters were replaced with `EnvironmentInterface`. I suppose they could be added to the baseline instead, and keep accessing public `Environment` properties as before if that's preferred.

Let me know of any issues, or any different approach preferable.

Once a proper solution is found i'll also document the changes.

### DIFF
--- a/src/Service/EnvironmentFactory.php
+++ b/src/Service/EnvironmentFactory.php
@@ -1,0 +1,16 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Laminas\Session\Service;
+
+use Laminas\Session\Validator\Environment;
+use Laminas\Session\Validator\EnvironmentInterface;
+
+final class EnvironmentFactory implements EnvironmentFactoryInterface
+{
+    public function getEnvironment(): EnvironmentInterface
+    {
+        return Environment::fromGlobals($_SERVER);
+    }
+}
--- a/src/Service/EnvironmentFactoryInterface.php
+++ b/src/Service/EnvironmentFactoryInterface.php
@@ -1,0 +1,12 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Laminas\Session\Service;
+
+use Laminas\Session\Validator\EnvironmentInterface;
+
+interface EnvironmentFactoryInterface
+{
+    public function getEnvironment(): EnvironmentInterface;
+}
--- a/src/Service/EnvironmentFactoryInterface.php
+++ b/src/Service/EnvironmentFactoryInterface.php
@@ -6,7 +6,12 @@ namespace Laminas\Session\Service;
 
 use Laminas\Session\Validator\EnvironmentInterface;
 
+/**
+ * Allows users to override the default {@see GlobalEnvironmentFactory} in DI
+ * in order to return custom {@see Environment} implementations
+ */
 interface EnvironmentFactoryInterface
 {
+    /** Retrieve an instance of {@see EnvironmentInterface} based on the configured factory */
     public function getEnvironment(): EnvironmentInterface;
 }
--- a/src/Service/SessionManagerFactory.php
+++ b/src/Service/SessionManagerFactory.php
@@ -14,12 +14,14 @@ use Laminas\Session\ManagerInterface;
 use Laminas\Session\SaveHandler\SaveHandlerInterface;
 use Laminas\Session\SessionManager;
 use Laminas\Session\Storage\StorageInterface;
+use Laminas\Session\Validator\EnvironmentInterface;
 use Psr\Container\ContainerInterface;
 
 use function array_merge;
 use function class_exists;
 use function get_debug_type;
 use function is_array;
+use function is_string;
 use function is_subclass_of;
 use function sprintf;
 
@@ -106,6 +108,8 @@ final class SessionManagerFactory implements FactoryInterface
             }
         }
 
+        $environmentClass = null;
+
         // Get session manager configuration, if any, and merge with default configuration
         if ($container->has('config')) {
             $configService = $container->get('config');
@@ -123,7 +127,16 @@ final class SessionManagerFactory implements FactoryInterface
             if (isset($managerConfig['options'])) {
                 $options = $managerConfig['options'];
             }
+
+            if (
+                isset($managerConfig['environment'])
+                && is_string($managerConfig['environment'])
+                && is_subclass_of($managerConfig['environment'], EnvironmentInterface::class)
+            ) {
+                $environmentClass = $managerConfig['environment'];
+            }
         }
+
         $managerClass = class_exists($requestedName) ? $requestedName : SessionManager::class;
         if (! is_subclass_of($managerClass, ManagerInterface::class)) {
             throw new ServiceNotCreatedException(sprintf(
@@ -133,7 +146,7 @@ final class SessionManagerFactory implements FactoryInterface
             ));
         }
 
-        $manager = new $managerClass($config, $storage, $saveHandler, $validators, $options);
+        $manager = new $managerClass($config, $storage, $saveHandler, $validators, $options, $environmentClass);
 
         // If configuration enables the session manager as the default manager for container
         // instances, do so.
--- a/src/Service/SessionManagerFactory.php
+++ b/src/Service/SessionManagerFactory.php
@@ -14,14 +14,12 @@ use Laminas\Session\ManagerInterface;
 use Laminas\Session\SaveHandler\SaveHandlerInterface;
 use Laminas\Session\SessionManager;
 use Laminas\Session\Storage\StorageInterface;
-use Laminas\Session\Validator\EnvironmentInterface;
 use Psr\Container\ContainerInterface;
 
 use function array_merge;
 use function class_exists;
 use function get_debug_type;
 use function is_array;
-use function is_string;
 use function is_subclass_of;
 use function sprintf;
 
@@ -65,12 +63,13 @@ final class SessionManagerFactory implements FactoryInterface
         string $requestedName,
         ?array $options = null
     ): ManagerInterface {
-        $config        = null;
-        $storage       = null;
-        $saveHandler   = null;
-        $validators    = [];
-        $managerConfig = $this->defaultManagerConfig;
-        $options       = [];
+        $environmentFactory = null;
+        $config             = null;
+        $storage            = null;
+        $saveHandler        = null;
+        $validators         = [];
+        $managerConfig      = $this->defaultManagerConfig;
+        $options            = [];
 
         if ($container->has(ConfigInterface::class)) {
             $config = $container->get(ConfigInterface::class);
@@ -108,8 +107,6 @@ final class SessionManagerFactory implements FactoryInterface
             }
         }
 
-        $environmentClass = null;
-
         // Get session manager configuration, if any, and merge with default configuration
         if ($container->has('config')) {
             $configService = $container->get('config');
@@ -127,13 +124,17 @@ final class SessionManagerFactory implements FactoryInterface
             if (isset($managerConfig['options'])) {
                 $options = $managerConfig['options'];
             }
+        }
 
-            if (
-                isset($managerConfig['environment'])
-                && is_string($managerConfig['environment'])
-                && is_subclass_of($managerConfig['environment'], EnvironmentInterface::class)
-            ) {
-                $environmentClass = $managerConfig['environment'];
+        if ($container->has(EnvironmentFactoryInterface::class)) {
+            $environmentFactory = $container->get(EnvironmentFactoryInterface::class);
+            if (! $environmentFactory instanceof EnvironmentFactoryInterface) {
+                throw new ServiceNotCreatedException(sprintf(
+                    'SessionManager requires that the %s service implement %s; received "%s"',
+                    EnvironmentFactoryInterface::class,
+                    EnvironmentFactoryInterface::class,
+                    get_debug_type($environmentFactory)
+                ));
             }
         }
 
@@ -146,7 +147,7 @@ final class SessionManagerFactory implements FactoryInterface
             ));
         }
 
-        $manager = new $managerClass($config, $storage, $saveHandler, $validators, $options, $environmentClass);
+        $manager = new $managerClass($config, $storage, $saveHandler, $validators, $options, $environmentFactory);
 
         // If configuration enables the session manager as the default manager for container
         // instances, do so.
--- a/src/SessionManager.php
+++ b/src/SessionManager.php
@@ -6,7 +6,8 @@ namespace Laminas\Session;
 
 use Laminas\EventManager\Event;
 use Laminas\EventManager\EventManagerInterface;
-use Laminas\Session\Validator\Environment;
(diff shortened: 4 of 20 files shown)


src/Validator/EnvironmentInterface.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Laminas\Session\Validator;
+
+interface EnvironmentInterface
+{
+    public static function fromGlobals(array $data): self;
```
gsteel:
```suggestion
    public static function fromGlobals(array $data): static;
```

src/SessionManager.php
```diff
@@ -84,6 +90,10 @@ public function __construct(
             $validators = array_merge($this->defaultValidators, $validators);
         }
 
+        if (is_string($environmentClass) && is_subclass_of($environmentClass, EnvironmentInterface::class)) {
+            $this->environmentClass = $environmentClass;
+        }
+
```
gsteel: Better to annotate the argument at `class-string<EnvironmentInterface>` than do these runtime checks IMO

src/Service/SessionManagerFactory.php
```diff
@@ -123,7 +127,16 @@ public function __invoke(
...
                 $options = $managerConfig['options'];
             }
+
+            if (
+                isset($managerConfig['environment'])
+                && is_string($managerConfig['environment'])
+                && is_subclass_of($managerConfig['environment'], EnvironmentInterface::class)
+            ) {
+                $environmentClass = $managerConfig['environment'];
+            }
```
gsteel: This config option will need documenting clearly.

It should also be added to the `ConfigProvider` along with necessary array shapes / psalm types etc.

src/SessionManager.php
gsteel: Also, is there a reason why this is not considered an option and included in the options array? This is a question not a suggestion!

src/SessionManager.php
```diff
@@ -62,6 +64,9 @@ final class SessionManager extends AbstractManager
 
     protected array $options = [];
 
+    /** @var class-string<EnvironmentInterface> */
+    protected string $environmentClass = Environment::class;
+
```
gsteel: Should be private.

Also, can we inline `AbstractManager` (Not here, but in another patch) - the inheritance looks mostly pointless and there are lots of opportunities to simplify this implementation. All the getters and setters should go, and the weird default config/storage class properties should go, all in favour of normal bog-standard DI.

src/SessionManager.php
```diff
@@ -169,25 +179,27 @@ protected function initializeValidatorChain(): void
         /** @var array<string, mixed> $storage */
         $storage = $this->getStorage()->getMetadata();
 
+        if (isset($storage['environment'])) {
+            assert(is_string($storage['environment']));
+            /** @var EnvironmentInterface $initialEnvironment */
+            $initialEnvironment = unserialize($storage['environment']);
+        } else {
+            $initialEnvironment = $this->environmentClass::fromGlobals($_SERVER);
```
gsteel: Still got `$_SERVER` here. I hate this.

gsteel (commented):
Not sure about passing a class name into session manager/session manager factory.

Maybe I'm over-complicating things way too much, but overall, if `SessionManager` required `EnvironmentFactory` in it's constructor, then fetching the current runtime environment would be reduced to `$this->envFactory->get()` (Or whatever) - removing superglobal access entirely.

At the moment, to test session manager, you still have to write to $_SERVER in order to correctly test all behaviour.

EnvironmentFactory could look like this:

```php
interface EnvironmentFactoryInterface {
    public function get(): EnvironmentInterface;
}

final class EnvironmentFactory implements EnvironmentFactoryInterface
{
   public function get(): EnvironmentInterface
   {
       return Environment::fromGlobals($_SERVER);
   }
}
```

This isolates globals usage to _one_, very simple place. Users override the factory in DI in order to return more specific environment implementations.

```php
final class UserEnvironmentFactory implements EnvironmentFactoryInterface
{
   public function get(): EnvironmentInterface
   {
       // Maybe I want to inspect headers sent by CloudFlare so I can implement validators around that
       return MySpecialEnvironment::fromGlobals($_SERVER);
   }
}
```

and config looks like

```php
return [
  'dependencies' => [
    'factories' => [
      EnvironmentFactory::class => InvokableFactory::class,
    ],
    'aliases' => [
      EnvironmentFactoryInterface::class => EnvironmentFactory::class,
    ],
  ],
];
```

I might not be the best person to review changes to this lib - I haven't used it since dropping MVC in favour of PSR-15 years ago.

Modernising/improving/simplifying this lib is a lot of work and I feel like I'm making it harder for you, potentially unnecessarily harder too.

What are the main (general) goals for the work you're trying to do here? and why?

src/SessionManager.php
Jurj-Bogdan: True, keeping the the current flow meant having to generate the  `Environment` after starting the session, and I've kept the "::fromGlobals" call here since it was only recently added.

The suggested factory itself as a dependency for `SessionManager` does indeed look like a better option

Jurj-Bogdan:
> What are the main (general) goals for the work you're trying to do here? and why?

I'm not sure i understand the question(s) - the current version of 3.0.x only has a single , final, `Environment` class containing only the specific 4 properties used by the included validators. 

As writing custom validators is [allowed (or is it to be removed?)](https://docs.laminas.dev/laminas-session/validators/writing-custom-validators/), that raises issues if users need other data from the `$_SERVER` superglobal.

The user's custom validators would have to use `Environment` objects to to satisfy the `ValidatorInterface`, but the missing data would have to be set "by hand" instead of making use of the `$initial` of `$current` objects, or simply skip implementing the shipped `ValidatorInterface` like the `Csrf` validator does.

The version in this PR would simply allow generating custom environments implementing `EnvironmentInterface`, allowing the `initial` and `current` objects to be used in whatever context the user needs.

src/SessionManager.php
gsteel: Given a factory that can produce an environment without any other arguments also means that `::fromGlobals()` can be dropped from the interface, leaving it present on the concrete class - custom environments/factories may produce an environment from something else, so forcing `fromGlobals` on them is not ideal

gsteel:
> I'm not sure i understand the question(s)

I was asking more broadly - i.e. not this patch, but all the work you have done/are doing to this lib

Jurj-Bogdan:
> > I'm not sure i understand the question(s)
> 
> I was asking more broadly - i.e. not this patch, but all the work you have done/are doing to this lib

Oh, in that case the original goal was simply having the library work with `laminas/laminas-servicemanager` 4, but as that requires a major version, more BC breaking changes were added in the various issues, PRs and their discussions.

If i'm not mistaken, the environment feature itself came about when adding types for issue #113, which in turn led to quite a few issues - with #133 leading to [PR 136](https://github.com/laminas/laminas-session/pull/136) that finally introduced the `Environment` usage.

gsteel:
It was my suggestion to introduce the value object, so wanted to validate that all of this was necessary.

For this lib to be relevant, it needs to be useful outside of MVC _(where, I assume it was traditionally used)_ becasue MVC is unlikely to get SMv4 support…

Jurj-Bogdan:
@gsteel Thanks for the feedback (will need more as well 🙃), added a factory for generating `Environment`s as it does indeed make a lot more sense and solves a few issues from the first version.

src/SessionManager.php
Jurj-Bogdan: Since it was already final i went ahead and made all properties private for now - the class will need a bigger rework away from the default properties usage, especially with a lot of tests making use of that flow.

src/SessionManager.php
Jurj-Bogdan: > Also, is there a reason why this is not considered an option and included in the options array? This is a question not a suggestion!

To be honest I think that's a remnant of an earlier version i tried where i was passing an object directly, which only replaced instead of moving to the `options` array, my bad

arhimede:
@laminas/technical-steering-committee can we have a review here ? 
In preparation for V3 , maybe we can discuss the release on next TSC meeting

src/Service/EnvironmentFactoryInterface.php
```diff
@@ -0,0 +1,12 @@
...
+
+declare(strict_types=1);
+
+namespace Laminas\Session\Service;
+
+use Laminas\Session\Validator\EnvironmentInterface;
+
+interface EnvironmentFactoryInterface
+{
+    public function getEnvironment(): EnvironmentInterface;
```
Ocramius, line 16: Would probably call this `GetEnvironment` or such? That's my opinion, though.

Also, let's document what this is for in the docblock

src/Validator/Environment.php
```diff
@@ -7,7 +7,7 @@
 use function is_string;
 use function session_id;
 
-final class Environment
+final class Environment implements EnvironmentInterface
 {
     public function __construct(
         public readonly ?string $userAgent = null,
```
Ocramius: let's maybe deprecate these public properties?

src/Service/EnvironmentFactory.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Laminas\Session\Service;
+
+use Laminas\Session\Validator\Environment;
+use Laminas\Session\Validator\EnvironmentInterface;
+
+final class EnvironmentFactory implements EnvironmentFactoryInterface
```
Ocramius: Let's rename this to `GlobalEnvironmentFactory`

src/SessionManager.php
```diff
@@ -169,25 +175,27 @@ protected function initializeValidatorChain(): void
         /** @var array<string, mixed> $storage */
         $storage = $this->getStorage()->getMetadata();
 
+        if (isset($storage['environment'])) {
+            assert(is_string($storage['environment']));
+            /** @var EnvironmentInterface $initialEnvironment */
+            $initialEnvironment = unserialize($storage['environment']);
+        } else {
+            $initialEnvironment = $this->environmentFactory->getEnvironment();
+            $this->getStorage()->setMetadata('environment', serialize($initialEnvironment));
```
Ocramius, line 184: We'd need to mark the `EnvironmentInterface` as serializable, I think

src/Validator/EnvironmentInterface.php
```diff
@@ -0,0 +1,16 @@
...
+
+interface EnvironmentInterface
+{
+    public function getUserAgent(): ?string;
+
+    public function getRemoteAddr(): ?string;
+
+    public function getForwardedFor(): ?string;
+
+    public function getSessionId(): ?string;
```
Ocramius, line 17: One problem comes to mind: you want to use custom `EnvironmentInterface` implementations, but then you end up with an environment that only supports these 4 fields? 

Wouldn't a key/value store be more useful then? :thinking: 

For example, if I were to write a session validator nowadays, I'd use something that accesses the entire `ServerRequestInterface` instance

src/Service/EnvironmentFactoryInterface.php
Jurj-Bogdan: Not sure i got this, renaming the interface to "GetEnvironmentFactoryInterface" or did you mean something else?

src/Validator/EnvironmentInterface.php
Jurj-Bogdan: These getters i was also unsure about, i've basically added them to force users to define the required properties for all shipped validators - even though only `Id` is set as default in the `SessionManager`.

I thought forcing the 4 original properties might be useful to avoid users getting errors with their custom validators in case they also add any combination of the shipped validators in their validator chains.

>  but then you end up with an environment that only supports these 4 fields?

This is somewhat related to the previous point, if the 4 getters aren't mandatory a user might create a `CustomEnvironment` that only holds the data he needs, but it won't be usable with any of the default validators if added.

The current version's downside is that the user has to set those properties even if he won't use them - let me know which version would be preferred.

> Wouldn't a key/value store be more useful then?

As in a `array $data` property instead of the current individual properties?

gsteel:
@Jurj-Bogdan Can't review today - please feel free to nag me to come back to this 👍

internalsystemerror (commented):
I may be a tad late at this, but I like the idea. Would it not be better though to have a ServerGlobalFactory, and then an EnvironmentGlobalFactory. The former loading from $_SERVER and the latter from $_ENV?

Maybe I'm missing the point of this PR.

Jurj-Bogdan:
> Maybe I'm missing the point of this PR.

The current version of `SessionManager` in 3.0.x basically [forces the usage](https://github.com/laminas/laminas-session/blob/51d3179353a11a1f507284aafef981e0fd2525a5/src/SessionManager.php#L187) of the shipped `Environment` which, while working with the 3 validators implementing `ValidatorInterface` in the package, would fail if a user would like to make use of the interface to add a custom validator using properties different than those found in the `Environment`.

This PR was made to allow users to create any custom validator that can still be chained and used by the `SessionManager::initializeValidatorChain` method

Jurj-Bogdan:
> @Jurj-Bogdan Can't review today - please feel free to nag me to come back to this 👍

@gsteel if you've got the time I could use another look

arhimede:
I am ready to merge this one in branch 3.x , in preparation for V3 release.
any other observation @laminas/technical-steering-committee  ?

Ocramius:
Thought about this some more myself: I'm not really happy with what the new interface restricts, but it's also an improvement over the previous approach (which was a fixed environment objecti).

Therefore, LGTM from my end.

arhimede (approved):

alexmerlin (approved):

arhimede:
thanks @Jurj-Bogdan

PR #163, opened 2025-10-24
Drop support for PHP 8.1 and add support for PHP 8.5, bump Psalm to 6.x and PHPUnit to 11.x

Jurj-Bogdan:
|    Q          |   A
|-------------- | ------
| RFC           | yes
| BC Break      | yes
| New Feature   | yes
| Version      | 3.0.x

### Description

Added PHP 8.5 support for the upcoming version 3.
Also updated PHPUnit to v11.x and Psalm to v6.x, which needed a few tweaks.

Thing is i have a few issues i could use some input about:

1. A new list of classes Psalm throws `ClassMustBeFinal` errors, which I've made final in this PR and will annotate as such in version 2.x if any remain final.

My question is if there are any of these you consider would be better of NOT final (maybe the `Storage` classes?), here's the full list:

```
Container
SessionConfig
BadMethodCallException
InvalidArgumentException
RuntimeException
Module
SessionArrayStorage
SessionStorage
ValidatorChain
```

2. The `xdebug` extension not being loaded in the PHPUnit 8.5 action - it's currently failing the phpunit tests for php 8.5 since xdebug 3.5 which supports it is still in alpha. I've tried it locally and the tests are passing, but i'm guessing requiring the alpha version in the workflow isn't allowed.

My question is if these tests that require the `xdebug` extension should be rewritten away from this dependency or just wait for the stable release xdebug 3.5.
Currently there are only 6 tests in the `SessionManagerTest` file which depend on this extension.

For now I haven't removed them as they seem quite useful, and using native functions like `headers_list` doesn't work in CLI. Also I'm not quite sure how to replicate the `xdebug_get_headers` function without weird changes in the actual code of `SessionManager`.

Obviously let me know of any other issues


arhimede:
@laminas/tsc-reviewers

Xerkus:
This needs to go into 2.x and then merged up into 3.0.x branch. Please rebase and adjust.

final can not be introduced there so that parts needs to be dropped. Introducing final keywords in a QA patch is too much for the scope so it would need to be done as a separate PR.

For now disable final check for src:
https://github.com/laminas/laminas-diactoros/blob/0ddf3f9ac5178731a3c9d0e648010700

Not sure what to do about xdebug in 8.5. Will need to check if there is any info on when it will be available in Ondrej ppa.

Xerkus:
xdebug support for php 8.5 is currently available in Xdebug 3.5.0alpha1

We might try to install it with pecl via pre-install script.

Jurj-Bogdan:
> xdebug support for php 8.5 is currently available in Xdebug 3.5.0alpha1
> 
> We might try to install it with pecl via pre-install script.



> xdebug support for php 8.5 is currently available in Xdebug 3.5.0alpha1
> 
> We might try to install it with pecl via pre-install script.

Thanks for the feedback - i'm pretty much done with the rebase, but i have one remaining issue with the xdebug extension.
From what i can see, neither [Ondrej ppa](https://launchpad.net/~ondrej/+archive/ubuntu/php/+index?batch=75&direction=backwards&start=225) nor [pecl](https://pecl.php.net/package/xdebug) have `xdebug 3.5alpha` available yet.

It seems to be available via `php-pie`, but i'm not sure if that's overkill or not - i'd have to install `pie`, then the extension via `pie` in order to make those 6 tests run, and this "hack" should also be removed whenever the extension is available normally.

 Should I go ahead with this approach or just skip the tests with something like `#[RequiresPhp('<8.5')]` until a major xdebug release is available?

gsteel:
> Should I go ahead with this approach or just skip the tests with something like `#[RequiresPhp('<8.5')]` until a major xdebug release is available?

IMO, we should wait for XDebug support in PHP 8.5, how ever long that takes. If we skip the tests now, they'll never get enabled again.

arhimede:
> > Should I go ahead with this approach or just skip the tests with something like `#[RequiresPhp('<8.5')]` until a major xdebug release is available?
> 
> IMO, we should wait for XDebug support in PHP 8.5, how ever long that takes. If we skip the tests now, they'll never get enabled again.

If we do not release now the PHP 8.5 support in this package, a lot of other packages and projects which depend on this are blocked.
I can personally guarantee that once xdebug will be release for php 8.5 , xdebug will be enabled back 😀

Let's release this onr without xdebug

froschdesign:
> Let's release this onr without xdebug

PHP 8.5 has not yet been released, so there is no _great_ rush.

arhimede:
for me it makes sense ...
@laminas/tsc-reviewers

Jurj-Bogdan:
For now I've just rebased the initial version intended for v3 to 2.26.x and committed the changes for review.

I'm working on a fix for the tests for php 8.4 right now, but I haven't touched the xdebug with php 8.5 issue however.

Jurj-Bogdan:
Rebased to 2.26.x to resolve conflicts and moved installing the xdebug and mongodb extensions to [php-pie](https://github.com/php/pie), as the latest versions needed for php 8.5 were not available via pecl or the default repo.

There's a few issues i'd appreciate some input with:
1. the failing psalm check i've for now left untouched: since [laminas-eventmanager](https://github.com/laminas/laminas-eventmanager/blob/3.15.0/src/EventManager.php#L22)'s new version marked the `EventManager` as soft final, i'm unsure how to go ahead with fixing the issue without breaking backward compatibility on this branch (2.26.x)
2.  when installing `PIE` i've skipped the `gh attestation` [check](https://github.com/laminas/laminas-session/blob/11f7c3451c628ee754fe9372463246d8b47fd995/.laminas-ci/install-mongodb-and-xdebug-extensions-via-pie.sh#L13) mentioned in their [one-liner installation guide](https://github.com/php/pie/blob/main/docs/usage.md#one-liner) as I'd also have to install the `github cli` which feels overkill for the action runners - is this alright?
3. I've run into an issue with the `mongodb/mongodb` dependecy when using php 8.5, as it was triggering phpunit errors, as the latest [stable release](https://github.com/mongodb/mongo-php-library/blob/2.1.1/src/functions.php#L412) contains PHP deprecations that for now are only fixed in the [dev version](https://github.com/mongodb/mongo-php-library/blob/v2.x/src/functions.php#L412). In order to skip this i've added a "hack" to [composer.json](https://github.com/laminas/laminas-session/blob/11f7c3451c628ee754fe9372463246d8b47fd995/composer.json#L50), i could use an opinion on how to avoid this

gsteel:
@Jurj-Bogdan 

If you can't fix a Psalm issue (Such as upstream deprecation or soft final) without causing a BC Break, just expand the baseline.

Before you run `psalm --set-baseline…` go through the errors to make sure none of them _are_ fixable first because once they're in the baseline they can easily be forgotten

Jurj-Bogdan:
Sure, if that's OK with you i can add those to the baseline - i though i might miss another solution though, that's why i mentioned it

gsteel:
> 3\. I've run into an issue with the `mongodb/mongodb` dependecy when using php 8.5, as it was triggering phpunit errors, as the latest [stable release](https://github.com/mongodb/mongo-php-library/blob/2.1.1/src/functions.php#L412) contains PHP deprecations that for now are only fixed in the [dev version](https://github.com/mongodb/mongo-php-library/blob/v2.x/src/functions.php#L412). In order to skip this i've added a "hack" to [composer.json](https://github.com/laminas/laminas-session/blob/11f7c3451c628ee754fe9372463246d8b47fd995/composer.json#L50), i could use an opinion on how to avoid this

Generally speaking, don't add dev branches in composer unless the work is WIP. In a library, you really shouldn't publish a release with a dependency on an unstable/unreleased dep because you are introducing potential BC breaks and instability directly into user projects.

You can ignore deprecations in 3rd party code in PHPUnit with `<source ignoreIndirectDeprecations="true">` in `phpunit.xml`

Jurj-Bogdan:
Added the Psalm issues coming from the soft final in `laminas-eventmanager` to the baseline for this version, to be refactored in v3.

Also added a separate commit where i removed the "dev version hack" to highlight the `mongodb/mongodb:2.1.1` errors mentioned earlier - i've been getting them even with `ignoreIndirectDeprecations="true"`.

Ocramius:
Wait, this was green, and went red again?

gsteel:
OK @Jurj-Bogdan - we still need to wait for a release: https://github.com/mongodb/mongo-php-library/issues/1795

Jurj-Bogdan:
> Wait, this was green, and went red again?

Yes, i've removed the "hack" i've added in a [previous commit](https://github.com/laminas/laminas-session/blob/ecdae15178d325a7f6ec07c92d23e642afded7b8/composer.json#L50) to better show the errors i was getting before adding the `dev` version of `mongodb/mongodb`.

gsteel:
`mongodb/mongodb` 2.1.2 is released: https://github.com/mongodb/mongo-php-library/releases/tag/2.1.2

Jurj-Bogdan:
Just bumped the minimum version to avoid issues with `php8.5 lowest`

psalm.xml
```diff
@@ -15,7 +15,34 @@
...
+        <UndefinedMagicPropertyFetch>
+            <errorLevel type="suppress">
+                <directory name="test" />
+            </errorLevel>
+        </UndefinedMagicPropertyFetch>
+        <UndefinedMagicMethod>
+            <errorLevel type="suppress">
+                <directory name="test" />
+            </errorLevel>
+        </UndefinedMagicMethod>
```
gsteel, line 44: If there is magic in tests, can it not be modernised and got rid of?

psalm.xml
```diff
@@ -15,7 +15,34 @@
             <directory name="vendor"/>
         </ignoreFiles>
     </projectFiles>
-
+    <issueHandlers>
+        <ClassMustBeFinal>
+            <errorLevel type="suppress">
+                <directory name="src" />
+            </errorLevel>
+        </ClassMustBeFinal>
```
gsteel, line 23: This is OK for now, but as work progresses on 3.x, soft final should be introduced in 2.x as a warning to users

src/AbstractContainer.php
```diff
@@ -477,11 +473,10 @@ public function exchangeArray($input)
...
         $old            = $storage[$name];
         $storage[$name] = $input;
-        if ($old instanceof ArrayObject) {
-            return $old->getArrayCopy();
-        }
 
-        return $old;
+        /** @var array<TKey, TValue> $return */
+        $return = $old instanceof ArrayObject ? $old->getArrayCopy() : $old;
+        return $return;
```
gsteel: You can avoid temporary variables in psalm with

```suggestion
        /** @psalm-var array<TKey, TValue> */
        return $old instanceof ArrayObject ? $old->getArrayCopy() : $old;
```

src/AbstractContainer.php
```diff
@@ -599,6 +594,8 @@ public function getArrayCopy()
         $storage   = $this->verifyNamespace();
         $container = $storage[$this->getName()];
 
-        return $container instanceof ArrayObject ? $container->getArrayCopy() : $container;
+        /** @var array<TKey, TValue> $array */
+        $array = $container instanceof ArrayObject ? $container->getArrayCopy() : $container;
+        return $array;
```
gsteel:
```suggestion
        /** @psalm-var array<TKey, TValue> */
        return $container instanceof ArrayObject ? $container->getArrayCopy() : $container;
```

arhimede:
finally :-) 

thanks @gsteel for the idea of pinging mongodb

src/SessionManager.php
```diff
@@ -97,24 +97,7 @@ public function __construct(
...
     public function sessionExists()
     {
-        if (session_status() === PHP_SESSION_ACTIVE) {
-            return true;
-        }
-
-        /**
-         * @var string|false $sid
-         */
-        $sid = defined('SID') ? constant('SID') : false;
```
gsteel, line 107: Removal of this might cause problems, see: https://github.com/laminas/laminas-session/pull/98

Can't remember exactly what the problem is, but it's of the legacy BC variety if memory serves. We may have to keep this in until 3.x

gsteel (changes requested):
Just a few things from me! Thanks for your work here @Jurj-Bogdan - it's taken a fair bit!!

psalm.xml
Jurj-Bogdan: There is indeed a lot of "magic" in the tests, mostly for the container (which at least for version 2 seems to be the [intended use case](https://docs.laminas.dev/laminas-session/v2/container/#basic-usage) ?) and storage tests (`StorageTest`, `SessionStorageTest`).

Since it's a valid use case i've kept the tests as they are for this version at least, maybe change it for v3 though. Let me know if that's not OK.

psalm.xml
Jurj-Bogdan: I'll mark the necessary files in a separate patch as per @Xerkus's earlier suggestion as well

Jurj-Bogdan:
@gsteel Thanks for the feedback!

I've made the requested changes, but kept a few instances of temporary variables used for assertions to silence psalm. like this fragment from `SessionConfig.php`:

```
$ret = ob_get_clean();
assert(is_string($ret));
return $ret;
```

Let me know if annotations are preferred though.

gsteel (approved):
Thanks again @Jurj-Bogdan LGTM 👍

PR #166, opened 2025-11-18
Merge release 2.26.0 into 3.0.x

Jurj-Bogdan:
### Description

I did not have permission to push to #165, i've made a separate branch for this update.

As discussed in #165, i've fixed the failing CI checks and will work of the other issues in separate patches (more final classes in `src/`, refactoring `ValidatorChain` away from the now soft final laminas-eventmanager `Event` class and probably more).


gsteel (approved):
Thanks @Jurj-Bogdan

Issue #167, opened 2025-11-19
[RFC]: Remove laminas/laminas-eventmanager dependency


Jurj-Bogdan:
## RFC

Removal of the `laminas/laminas-eventmanager` dependecy or refactoring that still includes it?

| Q                   | A      |
| ------------------- | ------ |
| Proposed Version(s) | 3.0.x  |
| BC Break?           | Yes |

### Description

Following the release of `laminas-eventmanager:3.1.5`, the `EventManager` class has been marked soft final. Currently it's being extended by [ValidatorChain](https://github.com/laminas/laminas-session/blob/3.0.x/src/ValidatorChain.php), which in turn is only used in the [SessionManager](https://github.com/laminas/laminas-session/blob/3.0.x/src/SessionManager.php#L173) and their respective tests. It doesn't seem to be used anywhere else.

In order not to extend the final class, what would you say the next step should be in version `3.0.x`? Refactoring the `ValidatorChain` (maybe using `EventManager` as a dependecy, or moving its features elsewhere?) while keeping the `laminas-eventmanager` dependency, or removing it altogether?

froschdesign:
I think it is completely irrelevant whether something is marked as final or not. Other questions need to be asked. Why is the event manager needed at all, and why is a chain for validators used with it? Why not rely on real validators instead and use the validator chain of laminas-validator?
Unfortunately, I don't know enough about this component to answer that question myself.

Jurj-Bogdan:
Maybe i haven't worded the issue properly - the issue with "final" was just found when updating [version 2](https://github.com/laminas/laminas-session/pull/163) where it was baselined, that's why i referenced it.

I don't think the event manager is mandatory in this new version though, but I could also use some feedback.

gsteel:
It seems that session validation is a core feature here, and, users can opt-in to a number of ready-made validators, and, create their own validators.

As far as I can tell, validation occurs at session startup, and there is a single validator chain that contains all the configured validators.

IMO there is no need to use EventManager to facilitate this setup. When SessionManager::isValid() is called, I'd just iterate over the composed validators and blow up as soon as one of them fails.

This may also provide an opportunity to simplify how the persistence and retrieval of the data necessary for session validation is handled, i.e. the before vs after comparisons that the validators rely on.

Considering validation failure is currently exceptional:

```php
// SessionManager.php

if (! $this->isValid()) {
    throw new Exception\RuntimeException('Session validation failed');
}
```

[...]

PR #168, opened 2025-11-22
Removed laminas-eventmanager dependency, refactored validators

Jurj-Bogdan:
|    Q          |   A
|-------------- | ------
| Documentation | no
| BC Break      | yes
| New Feature   | no
| RFC           | yes

### Description

Based on #167.

I agree with @gsteel 's suggestion of dropping the "event" flow for session validation, and since the dependency was not used in other places, I have removed `laminas/laminas-eventmanager`. To go in a bit more detail:

- the dependency was dropped.
- the `ValidatorChain` class that was using it has been removed.
  - as it was already agreed in previous PRs that going forward validators should only be added via config and only "automatically" called after session start, the whole storing of the configured validators under the "_VALID" key as previously done in `ValidatorChain` has also been completely removed.
- updated the shipped `Laminas\Session\ValidatorInterface` and all validators using it to suit the new validation flow.

While mentioned in the initial issue, I haven't touched the `Csrf` validator and its dependency `laminas/laminas-validator` here, as that's probably better suited for a new PR - right now this validator already cannot be used in the same way as the other validators implementing the own `ValidatorInterface` ( neither in v3 nor v2 ).


arhimede:
@laminas/tsc-reviewers 
Let's take a look at this one in order to have a release of version 3.0 until next Monday at TSC meeting

froschdesign:
> While mentioned in the initial issue, I haven't touched the `Csrf` validator and its dependency `laminas/laminas-validator` here, as that's probably better suited for a new PR

If this component defines its own interface for validators, then this should also be used by validators in this component. This not only reduces dependencies, but also confusion.

gsteel:
> If this component defines its own interface for validators, then this should also be used by validators in this component. This not only reduces dependencies, but also confusion.

The problem here is naming.

It's entirely aceptable to have 2 different types of validator:

- One for validating user input (via laminas-validator with the CSRF validator)
- One for validating sessions - as described here.

Renaming the session validator interface from `Session\ValidatorInterface` to `SessionValidator\SessionValidatorInterface` would reduce the potential for confusion at the expense of another BC break.

I'd support a rename of the interface/namespace as I find it unlikely that many users implement custom validators outside of those already shipped, and, the BC breaks are already very significant in this portion of the library.

The fact that the CSRF validator is in the same namespace is really unfortunate. Extracting it to a separate library (and removing the `laminas-validator` dependency) would clean things up nicely, but it's even more maintenance.

src/Validator/HttpUserAgent.php
```diff
@@ -4,33 +4,28 @@
...
      * Constructor
      * get the current user agent and store it in the session as 'valid data'
      */
-    public function __construct(
-        public readonly EnvironmentInterface $initial,
-        public readonly EnvironmentInterface $current,
-        array $option = []
-    ) {
+    public function __construct(array $options = [])
+    {
```
gsteel: The constructor should be dropped entirely here.

src/Validator/Id.php
```diff
@@ -18,23 +20,23 @@ final class Id implements ValidatorInterface
     /**
      * Constructor
      */
-    public function __construct(
-        public readonly EnvironmentInterface $initial,
-        public readonly EnvironmentInterface $current,
-        array $options = []
-    ) {
+    public function __construct(array $options = [])
+    {
```
gsteel: The constructor should be dropped entirely here.

Jurj-Bogdan:
> The fact that the CSRF validator is in the same namespace is really unfortunate. Extracting it to a separate library (and removing the `laminas-validator` dependency) would clean things up nicely, but it's even more maintenance.

@gsteel Should I maybe remove this feature altogether then? I'm not sure that form CSRF validation should be in the `laminas-session` library use case, especially not when it requires the `laminas-validator` dependency without much other use.

src/SessionManager.php
```diff
@@ -51,17 +50,14 @@ final class SessionManager extends AbstractManager
     private bool $sendExpireCookie;
     private bool $clearStorage;
 
-    /** @var list<class-string<ValidatorInterface>> $defaultValidators */
+    /** @var list<ValidatorInterface> */
     private array $defaultValidators = [
```
gsteel: This is still `list<class-string>`

gsteel: Also, if this were `public const DEFAULT_VALIDATORS = [...]`, then instead of mucking about with `class-string` anywhere in this class, all of it could be dealt with in the `SessionManagerFactory` so that `SessionManager` always gets `list<ValidatorInterface>` to its constructor. The benefit of that is the delegation of `new WhateverValidator()` to the service manager - each validator should be listed in `dependencies.factories` with `WhateverValidator::class => InvokableFactory::class` or a more specific factory to account for options arguments.

src/SessionManager.php
```diff
@@ -369,50 +336,35 @@ public function forgetMe(): static
...
+            assert(is_string($storage['environment']));
+            /** @var EnvironmentInterface $initialEnvironment */
+            $initialEnvironment = unserialize($storage['environment']);
+        } else {
+            $initialEnvironment = $this->environmentFactory->getEnvironment();
+            $this->getStorage()->setMetadata('environment', serialize($initialEnvironment));
+        }
+
+        foreach ($this->validators as $validatorName) {
+            $validator = new $validatorName($this->options);
```
gsteel: `newing` the validators here is very possibly a bad idea - we cannot guarantee constructor signatures because they are not subject to LS rules.

With `$this->validators` as a list of instances, provided in the constructor and newed in the factory via service manager, this problem is gone.

test/Validator/HttpUserAgentTest.php
```diff
@@ -14,38 +15,43 @@ final class HttpUserAgentTest extends TestCase
...
-
-        self::assertSame(HttpUserAgent::class, $validator->getName());
+        $validator = new HttpUserAgent();
+
+        $this->expectNotToPerformAssertions();
+        try {
+            $validator->validate(new Environment(userAgent: null), Environment::fromGlobals($_SERVER));
+        } catch (SessionValidationFailedException $e) {
+            $this->fail($e->getMessage());
+        }
```
gsteel: Better to `$this->expectException(…)`

test/Validator/HttpUserAgentTest.php
```diff
@@ -14,38 +15,43 @@ final class HttpUserAgentTest extends TestCase
...
-
-        self::assertTrue($validator->isValid());
+        $validator                  = new HttpUserAgent();
+
+        $this->expectNotToPerformAssertions();
+        try {
+            $validator->validate(Environment::fromGlobals($_SERVER), Environment::fromGlobals($_SERVER));
+        } catch (SessionValidationFailedException $e) {
+            $this->fail($e->getMessage());
+        }
```
gsteel: Remove the try/catch - the test will fail if an exception occurs

test/Validator/HttpUserAgentTest.php
```diff
@@ -14,38 +15,43 @@ final class HttpUserAgentTest extends TestCase
...
+
+        $this->expectNotToPerformAssertions();
+        try {
+            $validator->validate(
+                TestCustomEnvironment::fromGlobals($_SERVER),
+                TestCustomEnvironment::fromGlobals($_SERVER)
+            );
+        } catch (SessionValidationFailedException $e) {
+            $this->fail($e->getMessage());
+        }
```
gsteel: Unnecessary try/catch

test/Validator/IdTest.php
```diff
@@ -45,12 +46,25 @@ public function testIsValidPhp(int $bitsPerCharacter, string $id, bool $isValidP
     {
         ini_set('session.sid_bits_per_character', $bitsPerCharacter);
 
-        $validator = new Id(Environment::fromGlobals($_SERVER), new Environment(sessionId: $id));
-
+        $validator = new Id();
         if (PHP_VERSION_ID >= 80400) {
-            self::assertSame($isValidPhp84, $validator->isValid());
+            try {
```
gsteel: Same comments about try/catch here

test/Validator/IdTest.php
```diff
@@ -45,12 +46,25 @@ public function testIsValidPhp(int $bitsPerCharacter, string $id, bool $isValidP
     {
         ini_set('session.sid_bits_per_character', $bitsPerCharacter);
 
-        $validator = new Id(Environment::fromGlobals($_SERVER), new Environment(sessionId: $id));
-
+        $validator = new Id();
         if (PHP_VERSION_ID >= 80400) {
```
gsteel: Could we split this into 2 tests and use `#[RequiresPhp('>=8.4')]` attribute?

gsteel (changes requested):

test/Validator/HttpUserAgentTest.php
froschdesign: Reference: https://docs.phpunit.de/en/12.4/writing-tests-for-phpunit.html#expecting-exceptions

gsteel:
> @gsteel Should I maybe remove this feature altogether then? I'm not sure that form CSRF validation should be in the `laminas-session` library use case, especially not when it requires the `laminas-validator` dependency without much other use.

I agree, but I'd gather some more opinions first.

Who's going to make sure that there is a CSRF implementation available to users of Laminas Session for example. Is it even necessary based on the current fate of MVC? Mezzio already has an equivalent that doesn't use laminas-validator.

cc @laminas/technical-steering-committee

froschdesign:
> Is it even necessary based on the current fate of MVC?

I think the main use is in the [CSRF element of laminas-form](https://docs.laminas.dev/laminas-form/v3/element/csrf/).

gsteel:
> > Is it even necessary based on the current fate of MVC?
> 
> I think the main use is in the [CSRF element of laminas-form](https://docs.laminas.dev/laminas-form/v3/element/csrf/).

Ideally, `laminas-form` would not have hard or soft deps on `laminas-session`.

Maybe `laminas-csrf` _is_ necessary, so _it_ can have the dependency on both form and session.

On the other hand, why not just use mezzio-csrf 🤷‍♂️

gsteel:
@Jurj-Bogdan Perhaps it would be a good idea to bring this up as a topic for the next TSC meet.

1 - Is the CSRF validator and Form element relevant any more? Can they both be deprecated and removed?
2 - If not, should we be making a new library specifically for `laminas-csrf` to decouple `session` from `validator` and `form` from `session`

For the purposes of this patch, why don't we:

- Keep the existing naming (we can rename in a later patch depending on the outcome of the TSC meet)
- Retain the CSRF validator and leave it unchanged here (Awaiting TSC meet)

froschdesign:
> 1 - Is the CSRF validator and Form element relevant any more? Can they both be deprecated and removed?

The current usage of CSRF token generation and validation via laminas-form is easy:

```php
$form->add([
    'type' => Laminas\Form\Element\Csrf::class,
    'name' => 'example',
]);
```

Nothing else needs to be done. So this is very user-friendly.
I am in favour of decoupling, but user-friendliness must not suffer as a result. We already have this problem with mezzio-flash.

gsteel:
> Nothing else needs to be done. So this is very user-friendly.

Well, yes and no. Currently `form` is dependent on `validator@v2` which ships the _(deprecated)_ CSRF validator.

When `validator` is upgraded to v3 in `form`, the future of the CSRF element will be scrutinised because it automatically attaches the CSRF validator via `$element->getInputSpecification()` and it will need to depend on session to maintain the functionality of the csrf element.

For people using `mezzio` with `form`, all of this is irrelevant - they won't be using the csrf element, nor `laminas-session`, they'll be using `mezzio-session` and `mezzio-csrf`, so we'd be asking `mezzio` users to install `laminas-session` for functionality they don't want or need.

Should we be dogmatic about soft dependencies? I prefer to be, mainly because it places clear constraints on installable versions - with soft deps, we could end up with session:2 being installed with form:4 (This wouldn't happen because of validator:2 constraints, but, the point remains)

froschdesign:
@gsteel 
The technical details and dependencies are all clear, and I also do not want to have the laminas-session component in a Mezzio application. I am only concerned with user-friendliness! And mezzio-csrf simply is not user-friendly. Also mezzio-flash.
We can throw away the old elements, but we need replacements that are cleanly implemented and easy to use. So I do not want to stick to the current structures, just keep the user in mind.

Jurj-Bogdan:
While a decision is taken for the `Csrf` validator, i've gone through some of the requested changes. A couple of things:

- removed all unnecessary `try-catch` blocks and left phpunit to fail on the thrown exception where needed
  - the exception for this are the php <8.4 and >= 8.4 tests in `IdTest`, where the datasets specify if the validator should fail or not 
- added a new factory for the `RemoteAddr` validator so that it gets its `options` array passed, but I don't seem to find a required structure for its config - for now i've set it to `config.session_manager.options` based on test sequences:

[session_manager.validators](https://github.com/laminas/laminas-session/blob/2.27.x/test/Service/SessionManagerFactoryTest.php#L104)
[session_manager.options](https://github.com/laminas/laminas-session/blob/2.27.x/test/Service/SessionManagerFactoryTest.php#L215)

This might be better of reorganised, such as `session_manager.validators.{validator_name}.options` maybe?

gsteel:
Note https://github.com/laminas/technical-steering-committee/blob/main/meetings/minutes/2025-12-01-TSC-Minutes.md#discussion-around-v3-of-laminas-session

In a separate patch, we should delete anything to do with laminas-form or laminas-validator 👍

src/Service/RemoteAddressFactory.php
```diff
@@ -0,0 +1,35 @@
...
+
+/** @psalm-import-type OptionsArgument from RemoteAddr */
+final class RemoteAddressFactory implements FactoryInterface
+{
+    public function __invoke(
+        ContainerInterface $container,
+        string $requestedName,
+        ?array $options = null
+    ): RemoteAddr {
+        $config = $container->get('config');
```
gsteel: `$config` might be an `ArrayObject` or other iterable, so we should:

```php
$config = $container->has('config') ? $container->get('config') : [];
$config = is_iterable($config) ? iterator_to_array($config) : [];
// Now config is definitely an array…
```

gsteel: Also, `RemoteAddr` expects a specific shape - it's OK to force it here with

`@psalm-import-type Whatever` and `@psalm-var Whatever $whatever`

Is `session_manager.options` the correct key?

I'd expect something more specific to the `RemoteAddr` validator, such as `session_manager.remoteAddressOptions` or similar.

arhimede:
> Note https://github.com/laminas/technical-steering-committee/blob/main/meetings/minutes/2025-12-01-TSC-Minutes.md#discussion-around-v3-of-laminas-session
> 
> In a separate patch, we should delete anything to do with laminas-form or laminas-validator 👍

can you open an issue ? to remember later

src/SessionManager.php
```diff
@@ -369,50 +324,33 @@ public function forgetMe(): static
...
+        } else {
+            $initialEnvironment = $this->environmentFactory->getEnvironment();
+            $this->getStorage()->setMetadata('environment', serialize($initialEnvironment));
+        }
 
-        $responses = $validator->triggerEventUntil($falseResult, $event);
+        foreach ($this->validators as $validator) {
+            if (! $validator instanceof ValidatorInterface) {
+                continue;
+            }
```
gsteel: This condition is unnecessary because `$this->validators` is already `list<ValidatorInterface>`

src/SessionManager.php
```diff
@@ -369,50 +324,33 @@ public function forgetMe(): static
...
 
-        return ! $responses->stopped();
+            $validator->validate($initialEnvironment, $this->environmentFactory->getEnvironment());
```
gsteel: Possibly fetch the current environment in a variable outside of the loop. It's not doing any heavy lifting here, but other implementations might.

test/Service/SessionManagerFactoryTest.php
```diff
@@ -118,13 +130,12 @@ public function testFactoryWillAddValidatorViaConfiguration(): void
...
-
-        $listeners = iterator_to_array($this->getListenersForEvent('session.validate', $chain));
-
-        self::assertCount(2, $listeners);
+        $containedValidators = $this->getReflectionProperty($manager, 'validators');
+        self::assertIsArray($containedValidators);
+        self::assertCount(2, $containedValidators);
+        foreach ($containedValidators as $validator) {
+            self::assertTrue(is_a($validator, Validator\ValidatorInterface::class, true));
+        }
```
gsteel: Can we change this test so that:

- Create a manager with non-default validators
- Fix the environment so it will fail validation
- Assert validation fails

Reaching into private properties via reflection is brittle - better to test behaviour than code.

test/Service/SessionManagerFactoryTest.php
```diff
@@ -200,23 +206,17 @@ public function testFactoryDoesNotAttachValidatorTwoTimes(): void
...
-        }
-
-        $chain = $manager->getValidatorChain();
-
-        self::assertInstanceOf(EventManager::class, $chain);
-        $listeners = iterator_to_array($this->getListenersForEvent('session.validate', $chain));
-        self::assertCount(2, $listeners);
+        $containedValidators = $this->getReflectionProperty($manager, 'validators');
+        self::assertIsArray($containedValidators);
+        self::assertCount(2, $containedValidators);
```
gsteel: This test could be deleted - if a validator is attached more than once, behaviour would not change anyway, and we'd have one less test using reflection

gsteel (changes requested):
Sorry @Jurj-Bogdan! More nit-picks :)

src/Service/RemoteAddressFactory.php
```diff
@@ -0,0 +1,36 @@
...
+        if (! isset($config['session_manager'])) {
+            throw new ServiceNotCreatedException(
+                'Configuration is missing a "session_manager" key, or the value of that key is not an array'
+            );
+        }
+
+        $config = $config['session_manager'];
+        /** @var OptionsArgument $options */
+        $options = $config['remoteAddressOptions'] ?? [];
+        return new RemoteAddr($options);
```
gsteel:
```suggestion
        $sessionOptions = $config['session_manager'] ?? [];
        assert(is_array($sessionOptions));
        /** @psalm-var OptionsArgument $options */
        $options = $sessionOptions['remoteAddressOptions'] ?? [];
        return new RemoteAddr($options);
```

👆 That should clean up the baseline additions for this class

src/Service/RemoteAddressFactory.php
```diff
@@ -0,0 +1,36 @@
...
+
+use Laminas\ServiceManager\Exception\ServiceNotCreatedException;
+use Laminas\ServiceManager\Factory\FactoryInterface;
+use Laminas\Session\Validator\RemoteAddr;
+use Psr\Container\ContainerInterface;
+
+use function is_iterable;
+use function iterator_to_array;
+
+/** @psalm-import-type OptionsArgument from RemoteAddr */
```
gsteel:
```suggestion
/**
 * @internal
 *
 * @psalm-internal Laminas\Session
 * @psalm-internal LaminasTest\Session
 * @psalm-import-type OptionsArgument from RemoteAddr
 */
```

src/Service/SessionManagerFactory.php
```diff
@@ -14,15 +14,20 @@
...
+use function array_unique;
 use function class_exists;
 use function get_debug_type;
 use function is_array;
 use function is_subclass_of;
 use function sprintf;
 
+/**
+ * @psalm-import-type OptionsArgument from SessionManager
+ */
```
gsteel, line 36:
```suggestion
/**
 * @internal
 *
 * @psalm-internal Laminas\Session
 * @psalm-internal LaminasTest\Session
 * @psalm-import-type OptionsArgument from SessionManager
 */
```

test/SessionManagerTest.php
```diff
@@ -745,27 +724,19 @@ public function testForgetMeShouldSendCookieWithZeroTimestamp(): void
...
     #[RunInSeparateProcess]
     #[IgnoreDeprecations]
     public function testResumeSessionThatFailsAValidatorShouldRaiseException(): void
     {
-        $this->manager = new SessionManager(validators: [TestFailingValidator::class]);
-        $this->expectException(RuntimeException::class);
-        $this->expectExceptionMessage('failed');
+        $this->manager = new SessionManager(validators: [new TestFailingValidator()]);
+        $this->expectException(SessionValidationFailedException::class);
+        $this->expectExceptionMessage('Validation failed');
```
gsteel: This test is now identical to the previous test?

test/Validator/IdTest.php
```diff
@@ -38,62 +38,116 @@ public static function id(): iterable
...
+    ): void {
         ini_set('session.sid_bits_per_character', $bitsPerCharacter);
-
-        $validator = new Id(Environment::fromGlobals($_SERVER), new Environment(sessionId: $id));
-
-        if (PHP_VERSION_ID >= 80400) {
-            self::assertSame($isValidPhp84, $validator->isValid());
-        } else {
-            self::assertSame($isValidPhpPre84, $validator->isValid());
+        $validator = new Id();
```
gsteel, line 52:
```suggestion
        $validator = new Id();
        $isValid   = null;
```

to reduce baseline additions

test/Validator/IdTest.php
```diff
@@ -38,62 +38,116 @@ public static function id(): iterable
...
+    #[RunInSeparateProcess]
+    #[RequiresPhp('<8.4')]
+    public function testIsValidPhpPre84(
+        int $bitsPerCharacter,
+        string $id,
+        bool $isValidPhpPre84,
+        bool $isValidPhp84
+    ): void {
+        ini_set('session.sid_bits_per_character', $bitsPerCharacter);
+        $validator = new Id();
```
gsteel, line 75:
```suggestion
        $validator = new Id();
        $isValid   = null;
```

to reduce baseline additions

test/Validator/IdTest.php
```diff
@@ -38,62 +38,116 @@ public static function id(): iterable
...
-    public function testIsValidPhpWithCustomEnvironment(
+    #[RequiresPhp('>=8.4')]
+    public function testIsValidPhp84AndNewerWithCustomEnvironment(
         int $bitsPerCharacter,
         string $id,
         bool $isValidPhpPre84,
         bool $isValidPhp84
     ): void {
         ini_set('session.sid_bits_per_character', $bitsPerCharacter);
+        $validator = new Id();
```
gsteel, line 98:
```suggestion
        $validator = new Id();
        $isValid   = null;
```

test/Validator/IdTest.php
```diff
@@ -38,62 +38,116 @@ public static function id(): iterable
...
+    #[RunInSeparateProcess]
+    #[RequiresPhp('<8.4')]
+    public function testIsValidPhpPre84WithCustomEnvironment(
+        int $bitsPerCharacter,
+        string $id,
+        bool $isValidPhpPre84,
+        bool $isValidPhp84
+    ): void {
+        ini_set('session.sid_bits_per_character', $bitsPerCharacter);
+        $validator = new Id();
```
gsteel, line 127:
```suggestion
        $validator = new Id();
        $isValid   = null;
```

gsteel (changes requested):
Just a few tweaks to

- Reduce baseline additions
- Make factories internal to reduce the BC surface

Otherwise 👍

test/SessionManagerTest.php
Jurj-Bogdan: Indeed, since adding validators by hand is no longer allowed, the original test case is no longer relevant so i dropped it now.

src/Service/RemoteAddressFactory.php
```diff
@@ -0,0 +1,46 @@
...
+final class RemoteAddressFactory implements FactoryInterface
+{
+    public function __invoke(
+        ContainerInterface $container,
+        string $requestedName,
+        ?array $options = null
+    ): RemoteAddr {
+        $config = $container->has('config') ? $container->get('config') : [];
+        $config = is_iterable($config) ? iterator_to_array($config) : [];
+        if (! isset($config['session_manager'])) {
```
froschdesign: This is the factory for the validator. Why is the configuration searched under the key `session_manager`?

Jurj-Bogdan: Yes, i asked about that in an earlier commit since there doesn't seem to be a specific config structure defined. Initially I was thinking that since validators are now only added via the `SessionManager` i could go with something like "session_manager.validators.{validator_name}.options" or similar, but eventually opted for @gsteel's suggestion from an earlier review:

> Is `session_manager.options` the correct key?
> 
> I'd expect something more specific to the `RemoteAddr` validator, such as `session_manager.remoteAddressOptions` or similar.

froschdesign: > Initially I was thinking that since validators are now only added via the SessionManager…

The validators can be used independently without the `SessionManager`, so I see no need to use the `session_manager` configuration key.
But the validator doesn't implements the interface `Laminas\Session\Validator\ValidatorInterface` from laminas-validator, so it can't be used with the validator plugin manager, which means it can't be used the `validators` configuration key.
Therefore, I would expect `session` as key: `session.validators.…`.

gsteel: @froschdesign These are just not validators in the usual sense - they have nothing to do with laminas-validator. They are more like assertions that throw when something is "off" with a session

froschdesign: @gsteel
> These are just not validators in the usual sense - they have nothing to do with laminas-validator.

Correct.

> They are more like assertions that throw when something is "off" with a session.

So far, everything is clear.

The question is: can the validators be used independently? If yes, then I would expect `session` as key: `session.validators.…`. If the `SessionManager` is needed, then `session_manager.…`.

gsteel: I see what you mean - you wouldn't use these validators outside of laminas-session, and they are only called from the SessionManager context - in that regard, I have no preference on where the config is nested, but `session_manager` is probably a better target because it is at least laminas specific - ideally, the top level config key would be `laminas_session` to prevent potential conflicts with other software/libs, but that ship has sailed.

I propose `session_manager.validators.remote_addr` as the consistent place where any validators requiring configuration can reside, but I wouldn't die on that hill and think that `session.validators.remote_addr` would also be fine.

At this point I'm not sure which key is used to select the actual validators in use, i.e.

```php
return [
    'session_manager' => [
        'The validators I want to run' => [
            SomeValidator::class,
            OtherValidator::class,
        ],
    ],
];
```

Where does this config go?, because keeping them close together makes sense to me.

froschdesign: > Where does this config go?, because keeping them close together makes sense to me.

Makes also sense to me! 👍🏻

Jurj-Bogdan: @gsteel @froschdesign Sorry but I'm not sure i get the conclusion - would you prefer the remote_addr config under the `session_manager` or move it to a new `session.validators` key?

gsteel: > At this point I'm not sure which key is used to select the actual validators in use

Where does the user specify which validators to run?

Assuming this is `session.validators`, then config for the "RemoteAddr" validator should be nested under `session.validator_config.remote_addr` to keep these 2 items close to each other in the config.

Use naming that makes the most sense to users, i.e. don't use my suggestions if they are not appropriate for any reason.

Jurj-Bogdan: > Where does the user specify which validators to run?

It is under the `session_manager.validators` key, that's where the `SessionManagerFactory` looks for the validators to inject alongside the default validator specified in `SessionManager` itself.

gsteel: OK, so it would make sense to nest validator specific config under the same top-level key, so, as a suggestion `session_manager.validator_config.remote_addr` (Using snake/camel case as appropriate for consistency).

@froschdesign Anything to add?

froschdesign: This means two configurations are required:

1. which validators to run
2. the options for the different validators

```
session_manager.validators.execution = [
    RemoteAddr::class
]
session_manager.validators.options = [
    remote_addr = [],
    …
]
```

Correct?

Jurj-Bogdan: At the moment, yes, it would require a new key to hold the config... But as it's a feature for 3.0 onwards now would be the time to restructure the config to a clearer structure if you so desire.

Jurj-Bogdan: @froschdesign here's an updated config using your example as a starting point - i used the "classes" key instead of "execution" to keep it somewhat similar to the previous `session_manager.config.(class/options)` structure.

src/Service/RemoteAddressFactory.php
```diff
@@ -0,0 +1,48 @@
...
+            throw new ServiceNotCreatedException(
```
gsteel: As far as I can see, the `RemoteAddr` validator would be happy with `$options = []`, so throwing an exception here seems unnecessary. Also, the exception message does not match the conditional, there is no `! is_array`

src/Service/RemoteAddressFactory.php
```diff
@@ -0,0 +1,48 @@
...
+                'Configuration is missing a "session_manager" key, or the value of that key is not an array'
+            );
+        }
+        /** @psalm-var array<string, mixed> $sessionOptions */
+        $sessionOptions = $config['session_manager'];
```
gsteel:
```suggestion
        $sessionOptions = $config['session_manager'] ?? [];
        assert(is_array($sessionOptions));
```

src/Service/SessionManagerFactory.php
```diff
@@ -147,7 +161,36 @@ public function __invoke(
...
+        }
+
+        /** @psalm-var list<class-string<ValidatorInterface>> $uniqueValidators */
+        $uniqueValidators   = array_unique($validators);
+        $validatorInstances = [];
+        foreach ($uniqueValidators as $validator) {
+            if ($container->has($validator)) {
+                $validatorInstance = $container->get($validator);
+                if (! $validatorInstance instanceof ValidatorInterface) {
+                    throw new ServiceNotCreatedException(sprintf(
```
gsteel: This should be a `RuntimeException` from this lib IMO, not an exception type imported from SM - i.e. it's not our exception to throw. Not a hill I would die on though.

gsteel (changes requested):
Nearly there!!!

docs/book/v3/manager.md
```diff
@@ -38,8 +38,16 @@ return [
...
+            'classes' => [
+                Session\Validator\RemoteAddr::class,
+                Session\Validator\HttpUserAgent::class,
+            ],
+            'options' => [
+                'remote_addr' => [
+                    'use_proxy' => false,
+                    'trusted_proxies' => [],
+                ],
+            ]
```
Ocramius, line 50: This type is in the docs, but I don't directly see it reflected in any `@psalm-type` or array shape in this patch?

src/Service/SessionManagerFactory.php
```diff
@@ -41,10 +52,11 @@ final class SessionManagerFactory implements FactoryInterface
      * - Laminas\Session\Config\ConfigInterface
      * - Laminas\Session\Storage\StorageInterface
      * - Laminas\Session\SaveHandler\SaveHandlerInterface
+     * - Laminas\Session\Validator\ValidatorInterface
      *
-     * The first two have corresponding factories inside this namespace. The
-     * last, however, does not, due to the differences in implementations, and
-     * the fact that save handlers will often be written in userland. As such
+     * The first two have corresponding factories inside this namespace.
+     * "SaveHandlerInterface", however, does not, due to the differences in implementations,
```
Ocramius, line 58: Instead of a quoted string, use `{@see ...}`?

src/Service/SessionManagerFactory.php
```diff
@@ -118,7 +130,10 @@ public function __invoke(
             }
 
             if (isset($managerConfig['validators'])) {
-                $validators = $managerConfig['validators'];
+                $validatorsConfig = $managerConfig['validators'];
+                assert(is_array($validatorsConfig));
+                /** @var list<class-string<ValidatorInterface>> $validators */
+                $validators = $validatorsConfig['classes'];
```
Ocramius, line 136: This type is the one I flagged in the docs: kinda missing anywhere?

gsteel (approved):
Phew! Thanks @Jurj-Bogdan LGTM!

@froschdesign Anything to add?

froschdesign (approved):

src/Service/SessionManagerFactory.php
```diff
@@ -147,7 +162,36 @@ public function __invoke(
...
         }
 
-        $manager = new $managerClass($config, $storage, $saveHandler, $validators, $options, $environmentFactory);
+        /** @psalm-var OptionsArgument $options */
+        if ($options['attach_default_validators'] ?? true) {
+            $validators = array_merge(SessionManager::DEFAULT_VALIDATORS, $validators);
+        }
+
+        /** @psalm-var list<class-string<ValidatorInterface>> $uniqueValidators */
+        $uniqueValidators   = array_unique($validators);
```
Ocramius, line 171: Do we even care about uniqueness? What's the disadvantage if more than one is instantiated on accident?

src/Validator/Id.php
```diff
@@ -15,26 +17,19 @@
...
      */
-    public function isValid(): bool
+    public function validate(EnvironmentInterface $initial, EnvironmentInterface $current): void
     {
-        $sessionId = $this->current->getSessionId();
+        $sessionId = $current->getSessionId();
+
         if ($sessionId === null) {
-            return false;
+            throw new SessionValidationFailedException('Session id validation failed');
```
Ocramius, line 32: We should explain what failed (missing session ID)

src/Validator/Id.php
```diff
@@ -55,14 +50,8 @@ public function isValid(): bool
...
-    }
-
-    /**
-     * Return validator name
-     */
-    public function getName(): string
-    {
-        return self::class;
+        if (! (bool) preg_match($pattern, $sessionId)) {
+            throw new SessionValidationFailedException('Session id validation failed');
```
Ocramius, line 54: We should explain what failed (missing pattern mismatch - the ID and pattern should probably be part of the exception)

src/Validator/HttpUserAgent.php
```diff
@@ -4,33 +4,20 @@
...
-
-    /**
-     * Return validator name
-     */
-    public function getName(): string
+    public function validate(EnvironmentInterface $initial, EnvironmentInterface $current): void
     {
-        return self::class;
+        if ($initial->getUserAgent() !== $current->getUserAgent()) {
+            throw new SessionValidationFailedException('Http user agent validation failed');
```
Ocramius, line 20: The mismatching values should be part of the exception

src/Validator/RemoteAddr.php
```diff
@@ -29,27 +30,29 @@ final class RemoteAddr implements SessionValidator
...
+        if (isset($this->options['use_proxy']) && $this->options['use_proxy'] === true) {
+            $this->initialData = $this->getIpAddress($initial, $this->options);
+            $this->currentData = $this->getIpAddress($current, $this->options);
+        } else {
+            $this->initialData = $initial->getRemoteAddr();
+            $this->currentData = $current->getRemoteAddr();
+        }
+
+        if ($this->initialData !== $this->currentData) {
+            throw new SessionValidationFailedException('Remote address validation failed');
```
Ocramius, line 54: We should add the before/after values to the exception

src/Validator/ValidatorInterface.php
```diff
@@ -4,25 +4,13 @@
...
-     * that which was store in the setup() procedure.
-     */
-    public function isValid(): bool;
-
-    /**
-     * Get validator name for use with storing validators between requests
-     */
-    public function getName(): string;
+    /** @throws SessionValidationFailedException */
+    public function validate(EnvironmentInterface $initial, EnvironmentInterface $current): void;
```
Ocramius, line 15: This is obviously a major breaking change, but a welcome one 👍

Ocramius (commented):
From my PoV, this is a good/stateless design: I provided a lot of feedback on things to improve though, sorry :D

PR #169, opened 2025-12-05
removed unnecessary pre install scripts

Jurj-Bogdan:
|    Q          |   A
|-------------- | ------
| BC Break      |no
| RFC           | yes

### Description

With the release of the `xdebug:3.5.0` and the `mongodb:2.1.4` extensions there doesn't seem to be any reason to keep the pre-install scripts at all, as all extensions can now be installed as normal.


gsteel (approved):
Thanks @Jurj-Bogdan

gsteel:
It's not worth releasing this in the 2.x series right now, so feel free to create a merge-up PR for 3.0.x 👍

PR #170, opened 2025-12-05
remove unnecessary pre-install scripts

Jurj-Bogdan:
### Description

With the release of the xdebug:3.5.0 and the mongodb:2.1.4 extensions there doesn't seem to be any reason to keep the pre-install scripts at all, as all extensions can now be installed as normal.


gsteel (approved):
Thanks @Jurj-Bogdan 🙏

PR #174, opened 2026-02-20
Update v3 documentation, remove Module.php

Jurj-Bogdan:
|    Q          |   A
|-------------- | ------
| Documentation | yes
| New Feature   | no

### Description

Updating the documentation of version 3.


Jurj-Bogdan:
Will also amend version 2's migration documentation, as well as adding any missing `@final` or `@deprecated` annotations.

docs/book/v3/manager.md
```diff
@@ -53,8 +53,34 @@ return [
...
+        $this->defaultSessionManager->start();
+
+        return $handler->handle($request);
+    }
+}
+```
+
 The following illustrates how you might utilize the above configuration to
-create the session manager:
+create the session manager in a `laminas-mvc` application:
```
froschdesign: Version 3 of laminas-session can not be used in a laminas-mvc application. Any reference to this must therefore be removed.

docs/book/v3/validators/introduction.md
```diff
@@ -2,11 +2,16 @@
...
 - [Remote Addr](remoteaddr.md)
-- [Writing Custom Validators](writing-custom-validators.md)
 
-These validators are based on Laminas component for validation of data and files: [laminas-validator](https://docs.laminas.dev/laminas-validator/).
+The `Csrf` validator is based on Laminas component for validation of data and files: [laminas-validator](https://docs.laminas.dev/laminas-validator/).
+As such, its usage differs from the other shipped validators:
+
+- [Csrf](csrf.md)
 
 > MISSING: **Installation Requirements**
```
froschdesign: laminas-validator is a hard dependency in version 3, so this must be removed.

docs/book/v3/migration/v2-to-v3.md
```diff
@@ -42,6 +46,7 @@ In addition to these changes, there are several other important modifications:
 3. Options like `send_expire_cookie`, `clear_storage` and `preserve_storage` could be set during construction and will be used in `destroy()` and `start()` method calls if the parameter will be passed as null.
 4. All validator types are strictly defined as `class-string<ValidatorInterface>`.
 5. The implementation of the `sessionExists()` method has been simplified and no longer uses the PHP constant `SID` because this is [deprecated since PHP 8.4](https://wiki.php.net/rfc/deprecations_php_8_4#constant_sid).
+6. The `isValid` method no longer returns a boolean value, with validation failures throwing `SessionValidationFailedException`.
 
 > NOTE: **Logical change**
 > Due to this implementation change, `sessionExists()` will now return `false` after `session_close()` has been called, whereas in version 2 it would return `true`.
```
froschdesign:
```suggestion
NOTE: **Logical change**
Due to this implementation change, `sessionExists()` will now return `false` after `session_close()` has been called, whereas in version 2 it would return `true`.
```

froschdesign (changes requested):

docs/book/v3/validators/id.md
```diff
@@ -0,0 +1,8 @@
+# Session Id
```
froschdesign, line 1: When documentation files are added or removed, the configuration file for MkDocs, `mkdocs.yml`, must also be updated.

arhimede:
@laminas/tsc-reviewers 
Can we release the version 3 now ?

froschdesign:
The following is no longer needed in version 3:

https://github.com/laminas/laminas-session/blob/63e232f3ecfa699e9df0812fc20000464bad3ff3/mkdocs.yml#L47

froschdesign:
Some other/new configurations are not documented, for example:

https://github.com/laminas/laminas-session/blob/63e232f3ecfa699e9df0812fc20000464bad3ff3/src/Service/SessionManagerFactory.php#L136-L139

Jurj-Bogdan:
> The following is no longer needed in version 3:

Since its only this file and its tests, i've also removed them in this PR as well, hope its not too much of an issue.

arhimede:
@laminas/tsc-reviewers 
I am about to release the version 3 of this library, any other observation ?

docs/book/v3/manager.md
```diff
@@ -53,130 +53,28 @@ return [
...
-                default:
-                    $validator = new $validator();
-                    break;
-            }
-
-            $chain->attach('session.validate', array($validator, 'isValid'));
-        }
+    public function __construct(protected SessionManager $sessionManager) {
+        $this->defaultSessionManager = $sessionManager;
+        Container::setDefaultManager($sessionManager);
```
froschdesign: That looks confusing. Why does the standard manager have to be set in the middleware? Couldn't this have been solved using a generic factory or something similar, and the static stuff removed completely?
This part is not explained further in the documentation either, except that it would be a simple implementation.

Jurj-Bogdan: It was just intended as a simple illustration, considering the previous example was for `laminas-mvc` usage. Since that is no longer supported I simply added a quick example within a middleware (since this seems to be the intended use case for version 3).

froschdesign: The example creates more questions then it answers:

- What is this example actually supposed to show? It is not explained beforehand either.
- Why are there two properties? `$sessionManager` and `$defaultSessionManager`?
- Why is a default manager set to container here?

---

Regardless of the example, why is static setting in the container necessary at all? Can't we just get rid of it?
Do we still need this “old magic” anywhere? I don't think so!

Jurj-Bogdan: > Regardless of the example, why is static setting in the container necessary at all? Can't we just get rid of it? Do we still need this “old magic” anywhere? I don't think so!

My bad, that is very much true. Adjusted the example.

Also added a few extra bits of config information that could be useful for new users. I thought about adding a `session.global.php.dist` or similar file that would contain the config structure the package expects, but it doesn't seem it's a regular thing to do in other laminas packages. Instead i opted to at least mention the keys used by the "modules" in their documentation files, where they were missing.

docs/book/v3/manager.md
```diff
@@ -49,134 +49,39 @@ return [
                 ],
             ]
         ],
+        'options' => [
+            'preserve_storage' => false,
+            'send_expire_cookie' => true,
+            'clear_storage' => false,
+            'attach_default_validators' => true,
+        ],
```
froschdesign, line 50: Where are these options explained in detail?

docs/book/v3/manager.md
```diff
@@ -49,134 +49,39 @@ return [
...
-                            ?  $session['config']['class']
-                            : SessionConfig::class;
-
-                        $options = isset($session['config']['options'])
-                            ?  $session['config']['options']
-                            : [];
+    public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface {
+        $this->sessionManager->setId('customSessionId');
+        $this->sessionManager->setName('customSessionName');
+        $this->sessionManager->start();
```
froschdesign, line 90: Does this have to be done in middleware, or can it also be done via the configuration? The documentation hasn't mentioned this yet.

docs/book/v3/save-handler.md
```diff
@@ -12,6 +11,9 @@ PHP save handler when utilized in conjunction with
 handler. Generally if you are utilizing the `Cache` save handler; you are likely
 using products such as memcached.
 
+Note that `laminas-session` does **not** provide a `simple cache` implementation, and as such
+you will need to provide one of your choosing in order to use the shipped `Cache` save handler.
```
froschdesign: Turn the statement around and use a positive statement to say that there is a free choice, using the following example from laminas-cache.

docs/book/v3/storage.md
```diff
@@ -2,8 +2,20 @@
...
+
+```php
+    'session_storage' => [
+        'type' => SessionArrayStorage::class,
+        'options' => [
+            'input' => [],
+            'flags' => ArrayObject::ARRAY_AS_PROPS,
+            'iterator_class' => ArrayIterator::class,
+        ],
+    ],
```
froschdesign, line 17: Where are these options explained in detail?
It’s also confusing that, in the example, the type is set to the default value. It looks as though this needs to be specified. We’ve had this problem before, where users’ configurations were set to the default value.

froschdesign (changes requested):

docs/book/v3/manager.md
Jurj-Bogdan: The session name can indeed be set via config ( using the `StandardConfig` configuration array ), but the id itself cannot.

Speaking of the middleware example though, i am open to suggestions for a better alternative. I only wanted to illustrate a way to use the manager in whichever class would need to make use of its methods, not to specifically show how setting the id/name should be done.

mkdocs.yml
```diff
@@ -28,14 +28,13 @@ nav:
         - "Session Storage": v3/storage.md
         - Validators:
             - Introduction: v3/validators/introduction.md
+            - "Session Id": v3/validators/id.md
```
froschdesign: @Jurj-Bogdan 
In this file is a bug because the order of the versions in `nav` is wrong.

> # Versioned Contents
> 
> Documentation can be separated into different versions.
> Several steps are necessary for this.
> 
> ## Usage
> 
> 1. A separate subfolder is necessary for each version.
>     * `docs`
>       * `book`
>         * `v3`
>             * `introduction.md`
>             * `installation.md`
>             * …
>         * `v2`
>             * `introduction.md`
>             * `installation.md`
>             * …
>         * `index.md`
> 1. Extend the `nav` section of the MkDocs configuration file and a new section must be created for each version.
>     Add the related navigation entries in these sections.
>     ```yaml
>     nav:
>         - Home: index.md
>         - v3:
>             - Introduction: v3/introduction.md
>             - Installation: v3/installation.md
>             - …
>         - v2:
>             - Introduction: v2/introduction.md
>             - Installation: v2/installation.md
>             - …
>     ```
> 1. Extend the `extra` section of the MkDocs configuration file to add all available versions and the last / current version:
>     ```yaml
>     extra:
>         current_version: v3
>         versions:
>             - v3
>             - v2
>     ```
> 
> WARNING: **Order of Versions**
> The order of the versions in the `nav` and `versions` array is important.
> The newest version should be listed first, followed by older versions.
> 
> ## Example of Implementation
> 
> An example can be found in the [documentation of laminas-cache](https://github.com/laminas/laminas-cache/blob/070f1942d031abb8ce976e3cfdbd74a2674655c8/mkdocs.yml#L3-L48).
>

froschdesign (changes requested):

froschdesign (approved):

arhimede:
thanks @Jurj-Bogdan

PR #178, opened 2026-04-03
Removed headers_sent check from `SessionManager::sessionExists`

### LINKED ISSUE #123: `SessionManager::sessionExists()` considers `headers_sent() === true` as evidence of session
### Bug Report

|    Q        |   A
|------------ | ------
| Version(s)  | `2.16.0`

#### Summary

I'm not quite the original motivation but when `::sessionExists()` looks for evidence of a session, it considers `headers_sent() === true` as sufficient proof... except there are plenty of valid reasons for having output already sent to the client or console, e.g. during PHPUnit tests.

#### Current behavior

`headers_sent()` is evidence of a session being present, probably for very old Internet reasons.

#### How to reproduce

Run a test with PHPUnit (not run in an isolated environment - yes this is a workaround but not a good one) and `headers_sent()` will always be `true`, which means the session manager goes astray thinking we already have an active session.

#### Expected behavior

We remove this check and allow the other, less flaky, tests remain.

Jurj-Bogdan:
|    Q          |   A
|-------------- | ------
| Bugfix        | yes
| BC Break      | no
| Target version | 2.17

### Description

Based on issue #123 : removed the `headers_sent` check from the `SessionManager::sessionExists` method, which would turn the `session_start` errors into silent failures if output was emitted for any reason before starting the session.

Without this check, the error would be thrown in order for the user to be able to investigate it.

### DIFF
--- a/src/SessionManager.php
+++ b/src/SessionManager.php
@@ -112,10 +112,6 @@ class SessionManager extends AbstractManager
             return true;
         }
 
-        if (headers_sent()) {
-            return true;
-        }
-
         return false;
     }
 


src/SessionManager.php
froschdesign, line 115: Can this be tested? Something is being removed here without any test indicating this, so something is missing.

arhimede:
thanks @Jurj-Bogdan

arhimede (approved):

PR #179, opened 2026-04-03
Removed headers_sent check (v3)

Jurj-Bogdan:
|    Q          |   A
|-------------- | ------
| Bugfix        | yes
| BC Break      | no
| Target version | 3.0.x

### Description

Based on issue #123 : removed the `headers_sent` check from the `SessionManager::sessionExists` method, which would turn the `session_start` errors into silent failures if output was emitted for any reason before starting the session.

Without this check, the error would be thrown in order for the user to be able to investigate it.

### DIFF
--- a/src/SessionManager.php
+++ b/src/SessionManager.php
@@ -88,7 +88,7 @@ final class SessionManager extends AbstractManager
      */
     public function sessionExists(): bool
     {
-        return session_status() === PHP_SESSION_ACTIVE || headers_sent();
+        return session_status() === PHP_SESSION_ACTIVE;
     }
 
     /**


src/SessionManager.php
```diff
@@ -88,7 +88,7 @@ public function __construct(
      */
     public function sessionExists(): bool
     {
-        return session_status() === PHP_SESSION_ACTIVE || headers_sent();
+        return session_status() === PHP_SESSION_ACTIVE;
```
froschdesign, line 91: Can this be tested? Something is being removed here without any test indicating this, so something is missing.

Jurj-Bogdan: The `sessionExists` method itself was already covered in a few explicit tests from `SessionManagerTest` (and it is being called on every `SessionManager::start()`, so it's covered there as well). Since the method's role is unchanged, i thought they should be enough.

However, as mentioned in the original issue itself, I'm not even sure why a `headers_sent` check would be needed any more. The `sessionExists` method should only answer if there is an active session - in this case, the `headers_sent` check could return false positives for cases where output is legitimately sent - or alternatively would supress errors where unwanted output is sent before the session start instead of allowing them to be caught.

I might be missing something of course, feel free to let me know.

samsonasik: @Jurj-Bogdan failing test case is needed if there is a bug, that show if that rolled back, the bug will be shown

Jurj-Bogdan: @samsonasik  Alright, provided i'm not misunderstanding the request, a test that fails if the `headers_sent` call is reintroduced in the `SessionManager::sessionExists` method is required, right?

In that case, i'm not sure how exactly I could test that considering no headers are sent in cli - i could maybe [edit this test for example](https://github.com/laminas/laminas-session/blob/3.0.x/test/SessionManagerTest.php#L292) to show the intended functionality, like adding `self::assertFalse($this->manager->sessionExists());` after the `manager->start()` call? It would basically only show the intent though, as it will be skipped when using the cli sapi.

samsonasik: You may can create some custom function `header_sent()` in the unit test.

Actually, the git blame probably needed why it is here in the first place, and why marking session exists on header sent.

Jurj-Bogdan: >the git blame probably needed why it is here in the first place

It has been ported over from version 2, which in turn included this call from the [very first commit](https://github.com/laminas/laminas-session/blob/ce63e39edc03a3ec2c5f08a4ab82261dee7e5293/src/SessionManager.php#L56) of laminas-session. I also have a PR to remove it from version 2 as well, but I'll update that once a fix is found here.

> You may can create some custom function `header_sent()` in the unit test.

@samsonasik I have tried implementing this locally, but it feels very "hacky" for such a small change; as a summary:

- I can't override the `headers_sent()` in the test itself as the namespace won't match when trying to use it within `SessionManager`. i needed to add a separate "TestOverrides.php" file under the `Laminas\Session` namespace so that it would be picked up in `SessionManager::sessionExists()`;
  - in order not to affect the other tests, the custom `headers_sent()` function would default to returning `false` except when explicitly told to return `true` for this custom test only;
- this file would need to be added to PHPUnit's bootstrap, which already uses [autoload.php](https://github.com/laminas/laminas-session/blob/3.0.x/phpunit.xml.dist#L5). So in order to add this second file to PHPUnit's bootstrap, i needed another file, `boostrap.php` which would include both the original `vendor/autoload.php` and the previously mentioned `TestOverrides.php`. Then i had to set this as the phpunit bootstrap file;
- even with this hack, using the explicit `use function` [in SessionManager](https://github.com/laminas/laminas-session/blob/3.0.x/src/SessionManager.php#L16) would select the global `headers_sent` function, so i needed to remove this line (which also brings a phpcs "ReferenceViaFallbackGlobalName" error).

Basically, only after all these hacks would the new test make use of the custom "headers_sent" function when calling "SessionManager::sessionExists". The following test would pass without the `|| headers_sent()` and fail if it is added again:


```php
    public function testSessionExistsReturnsFalseEvenWhenHeadersAreSent(): void
    {
        try {
            self::$headersSent = true;
            $this->manager     = new SessionManager();

            self::assertFalse($this->manager->sessionExists());
        } finally {
            self::$headersSent = false;
        }
    }
```

Sorry, maybe i'm missing something obvious, but if not, adding all these hacks just for the "headers_sent" removal seems like overkill to me.

froschdesign: Thank you for your analysis.
How is that part handled within [mezzio-session-ext](https://github.com/mezzio/)?

froschdesign: @Xerkus 
Can you help with this?

Jurj-Bogdan: > Thank you for your analysis. How is that part handled within [mezzio-session-ext](https://github.com/mezzio/)?

To be honest i'm not familiar with mezzio-session(-ext), but from what i can see, it makes use of "session_status()" only in order to check if a session exists.

arhimede: > @Xerkus Can you help with this?

I am not sure if @Xerkus can reply soon ...

in `mezzio-session-ext` the problem is addressed the same like here 

```php
       if (session_status() === PHP_SESSION_ACTIVE) {
            $_SESSION = $session->toArray();
            session_write_close();
        }
```

Let's merge this PR and allow me to open a discussion about this major release, according to latest TSC decision . 
and move on  :-)

froschdesign: @arhimede 
It’s up to you. I just wanted to point out an inconsistency and a possible gap in the tests. My comment was made purely from a quality assurance perspective.

arhimede (approved):

arhimede:
thanks @Jurj-Bogdan

PR #184, opened 2026-07-07
[RFC] Lazy session start

### LINKED ISSUE #21: Lazy session start
The current container object starts a session in the constructor automatically. This works fine for most use cases, but it could come in handy to use some kind of delayed session start. I'm working on a REST API which accepts API keys, but also check for authenticated users in their sessions. The authentication for the users are stored in sessions. However, every API call from a REST client generates a session. It will send a cookie to the REST client, while also storing the session information on disk. This should not happen with every API call by REST clients.

It would be great to have a container which does not automatically starts a session, only under the following circumstances:
- If data is written to the container, start a new session
- If trying to read data from the container, only start a session when `session_status` return `PHP_SESSION_NONE` and `$_COOKIE[session_name()]` is set. This will only start a session when the browser actually has send a cookie to the server.

---

Originally posted by @dekker-m at https://github.com/zendframework/zend-session/issues/43

Jurj-Bogdan:
|    Q          |   A
|-------------- | ------
| RFC           | yes
| New Feature   | yes
| Documentation | yes

### Description

RFC for "lazy session container" feature mentioned in issue #21 .

To be honest i am not entirely sure what the requirements for this feature are - should it simply provide a way to not start the session when using `AbstractContainer` implementations? Or it the goal to implement a similar feature to `mezzio-session's LazySession`? 

The `mezzio-session` example provides a way to persist data in memory before starting the session, something that is not available in this package. This would require adding this mechanism from scratch (or bring in ` mezzio-session-ext` to this repo as well), and to be honest i'm not sure what the use case here would be when `mezzio-session` already covers the "Middleware-first approach".

For the purposes of discussing this, i went with the first option, with adding a new `LazyContainer` instead of changing the original `Container`. This new container simply defers starting the session when reading, as opposed to the `Container` calling `AbstractContainer::__construct` which automatically starts the session. Writing will always start the session, as per the issue's first point. 
Also adapted `ContainerAbstractServiceFactory` to allow instantiating`LazyContainer` instances, like the original `Container` ones (which are kept as the default `AbstractContainer` implementation used).

I could use more feedback about what exactly a "lazy container" should do in this project.

Since this is mostly a "proof of concept" implementation, the tests suite is a bit basic (duplicated `Container's` tests and added specific read/write session starting checks only). I'm also aware of the Psalm issues, which i'll of course resolve once a way forward is agreed upon, as well as expanding the test suite.

### DIFF
--- a/docs/book/v3/container.md
+++ b/docs/book/v3/container.md
@@ -18,6 +18,12 @@ $container = new Container('namespace');
 $container->item = 'foo';
 ```
 
+## Lazy Container
+
+`Laminas\Session\LazyContainer` is another `Laminas\Session\AbstractContainer` implementation,
+which defers starting the session until it's needed: reading does not start it unless a session cookie is already present,
+while writing always does.
+
 ## Setting the Default Session Manager
 
 In the event you are using multiple session managers or prefer to be explicit,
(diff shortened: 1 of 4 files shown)


arhimede:
@laminas/tsc-reviewers Can you take a look at this one ?
it adresses an 10 years old issue :-)

